PR #97, opened 2021-11-18
Use swift-install

adam-fowler:


adam-fowler:
It ran but mosquitto server is lost again.

Issue #101, opened 2022-02-01
Xcode preview fails to build with errors


drmarkpowell:
**Describe the bug**
Adding mqtt-nio dependency to iOS project breaks SwiftUI Previews

**To Reproduce**
1. Create a new iOS app project in Xcode
2. Add mqtt-nio dependency in Swift packages to the app project
3. open ContentView.swift and refresh the Preview. This generates 11 errors :

Undefined symbol: nominal type descriptor for NIOSSL.NIOSSLCertificate
Undefined symbol: nominal type descriptor for NIOSSL.NIOSSLClientTLSProvider
Undefined symbol: NIOSSL.NIOSSLCertificate.toDERBytes() throws -> [Swift.UInt8]
Undefined symbol: static NIOSSL.NIOSSLCertificate.fromPEMFile(Swift.String) throws -> [NIOSSL.NIOSSLCertificate]
Undefined symbol: type metadata accessor for NIOSSL.NIOSSLContext
Undefined symbol: nominal type descriptor for NIOSSL.TLSConfiguration
Undefined symbol: protocol conformance descriptor for NIOSSL.NIOSSLClientTLSProvider<A> : NIOCore.NIOClientTLSProvider in NIOSSL
Undefined symbol: type metadata accessor for NIOSSL.NIOSSLCertificate
Undefined symbol: NIOSSL.NIOSSLContext.__allocating_init(configuration: NIOSSL.TLSConfiguration) throws -> NIOSSL.NIOSSLContext
Undefined symbol: static NIOSSL.TLSConfiguration.makeClientConfiguration() -> NIOSSL.TLSConfiguration
Undefined symbol: NIOSSL.NIOSSLClientTLSProvider.init(context: NIOSSL.NIOSSLContext, serverHostname: Swift.String?, customVerificationCallback: (([NIOSSL.NIOSSLCertificate], NIOCore.EventLoopPromise<NIOSSL.NIOSSLVerificationResult>) -> ())?) throws -> NIOSSL.NIOSSLClientTLSProvider<A>

**Expected behavior**
mqtt-nio builds and runs fine in the iOS simulator and on a real iOS device but breaks previews in Xcode...the preview simulator fails with errors on the build.

**Context (please complete the following information):**
 - Platform [e.g. iOS, macOS, Linux] iOS
 - Swift/Xcode Version: Xcode 13.2.1
 - Version [e.g. 1.2.0] mqtt-nio latest 2.5.2

**Additional context**
[...]

PR #102, opened 2022-02-09
Only use NIOSSL on macOS and Linux (#101)

obriensp:
When building for SwiftUI Previews, checking if `NIOSSL` is available will succeed, despite the package has been configured to only link on macOS and Linux. This results in link errors which break SwiftUI Previews.

### DIFF
--- a/Sources/MQTTNIO/MQTTClient.swift
+++ b/Sources/MQTTNIO/MQTTClient.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the MQTTNIO project
 //
-// Copyright (c) 2020-2021 Adam Fowler
+// Copyright (c) 2020-2022 Adam Fowler
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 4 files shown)


adam-fowler:
Sorry been a bit slow to reply (I'm on holiday, until Monday). Your changes look fine although you shouldn't need the `canImport` if you add the platform checks. 

Also I would like to see the actual error myself before merging this though.

obriensp:
@adam-fowler No problem, and I hope you're having a great holiday!

For when you're back: I'm able to reproduce the issue using the EmCuTeeTee project in Xcode 13.2. I'll go ahead and update the conditionals to remove the `canImport`.

adam-fowler (changes requested):
Hi, one other thing can you stick a comment above each change of the form "This should use canImport(NIOSSL), will change when it works with SwiftUI previews". 

Ignore the validity check, there is a problem with file headers and swift format which needs resolved

Otherwise LGTM

adam-fowler (approved):
LGTM

PR #117, opened 2022-05-30
Add SPI DocC config

finestructure:


Package.swift
```diff
@@ -27,3 +27,11 @@ let package = Package(
...
     ]
 )
+
+
+#if swift(>=5.6)
+  // Add the documentation compiler plugin if possible
+  package.dependencies.append(
+    .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0")
+  )
+#endif
```
adam-fowler, line 37: I've been building docc outside of the package, this is a nice way to include it as a plugin.

adam-fowler (approved):
LGTM

PR #122, opened 2022-08-29
Use swift-atomics

adam-fowler:


Sources/MQTTNIO/MQTTClient.swift
```diff
@@ -11,6 +11,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Atomics
```
0xTim, line 14: I think you can remove `NIOConcurrencyHelpers` now right?

adam-fowler: I'm using `Lock` so I can't. There are a couple of other files though where I can remove it

PR #131, opened 2023-01-05
Fix re-entrancy issue with MQTTListeners.notify

adam-fowler:
Copy listener array when inside lock, process copied listener array outside lock

### DIFF
--- a/Sources/MQTTNIO/MQTTListeners.swift
+++ b/Sources/MQTTNIO/MQTTListeners.swift
@@ -14,14 +14,15 @@
 import NIO
 import NIOConcurrencyHelpers
 
-class MQTTListeners<ReturnType> {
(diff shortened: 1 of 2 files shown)


Sources/MQTTNIO/MQTTListeners.swift
```diff
@@ -38,9 +39,12 @@ class MQTTListeners<ReturnType> {
     }
 
     func removeAll() {
-        self.listeners = [:]
+        self.lock.withLock {
+            self.listeners = [:]
+        }
     }
 
+    // private let eventLoop: EventLoop
```
fabianfett, line 47: comment left over?

fabianfett (approved):
looks great!

PR #138, opened 2023-06-04
Remove deprecated LinuxMain

michalsrutek:
Since [Swift 5.4](https://www.swift.org/blog/swift-5.4-released/), there's automatic test discovery, removing the need for `LinuxMain.swift `.


adam-fowler:
Thank you

PR #146, opened 2023-11-19
Support Unix Domain Socket connections

jpgrayson:
The approach used for setting-up an `MQTTClient` to use a unix socket is similar to the approach used by mosquitto to configure listening on a unix socket; namely, setting the port to 0 indicates that the `host` string should be interpreted as a path to a socket.

From a semver perspective, this approach enables the unix domain sockets feature without breaking or otherwise changing an MQTTNIO interfaces. Prior to this change, specifying port 0 would have been an error, now it has meaning.

The docker-compose.yml files are updated to setup a shared "mosquitto-socket" volume. The mosquitto container will listen on a socket file in this volume. The "test"/"app" container also mounts this volume and thus can use the socket file found there to connect to mosquitto. The new unix socket configuration in the example mosquitto.conf is setup such that the socket path will work both with the docker-compose.yml setups as well as with a local (non-container) mosquitto instance.

The `mosquitto/socket` directory must be present in order to run a local mosquitto service with `mosquitto/conf/mosquitto.conf` since that configuration will listen on a socket in that directory. The `readme.md`
in that directory ensures its existence in git workspaces.

### DIFF
--- a/Sources/MQTTNIO/MQTTConnection.swift
+++ b/Sources/MQTTNIO/MQTTConnection.swift
@@ -46,7 +46,6 @@ final class MQTTConnection {
         do {
             // get bootstrap based off what eventloop we are running on
             let bootstrap = try getBootstrap(client: client)
-            bootstrap
                 .channelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
                 .channelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: 1)
                 .connectTimeout(client.configuration.connectTimeout)
@@ -75,7 +74,16 @@ final class MQTTConnection {
                         return channel.pipeline.addHandlers(handlers)
                     }
                 }
-                .connect(host: client.host, port: client.port)
+
+            let channelFuture: EventLoopFuture<Channel>
+
+            if client.port == 0 {
+                channelFuture = bootstrap.connect(unixDomainSocketPath: client.host)
+            } else {
+                channelFuture = bootstrap.connect(host: client.host, port: client.port)
+            }
+
+            channelFuture
                 .map { channel in
                     if !client.configuration.useWebSockets {
                         channelPromise.succeed(channel)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-# run this with docker-compose -f docker/docker-compose.yml run test
+# run this with: docker-compose -f docker-compose.yml run test
 version: "3.3"
 
 services:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,7 @@ services:
     working_dir: /mqtt-nio
     volumes:
       - .:/mqtt-nio
+      - mosquitto-socket:/mqtt-nio/mosquitto/socket
     depends_on:
       - mosquitto
     environment:
@@ -19,8 +20,12 @@ services:
     volumes:
       - ./mosquitto/config:/mosquitto/config
       - ./mosquitto/certs:/mosquitto/certs
+      - mosquitto-socket:/mosquitto/socket
     ports:
       - "1883:1883"
       - "8883:8883"
       - "8080:8080"
       - "8081:8081"
+
+volumes:
+  mosquitto-socket:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     environment:
       - MOSQUITTO_SERVER=mosquitto
       - CI=true
-    command: /bin/bash -xcl "swift test --enable-test-discovery"
+    command: /bin/bash -xcl "swift test"
 
   mosquitto:
     image: eclipse-mosquitto
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -6,6 +6,7 @@ services:
     image: swift:5.9
     volumes:
       - ..:/workspace
+      - mosquitto-socket:/workspace/mosquitto/socket
     depends_on:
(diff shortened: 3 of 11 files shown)


docker-compose.yml
```diff
@@ -7,20 +7,25 @@ services:
...
       - ./mosquitto/certs:/mosquitto/certs
+      - mosquitto-socket:/mosquitto/socket
     ports:
       - "1883:1883"
       - "8883:8883"
       - "8080:8080"
       - "8081:8081"
+
+volumes:
+  mosquitto-socket:
```
adam-fowler, line 31: Any reason we need the mosquitto-socket volume here?
It is only accessed from the `mosquitto`  service, I know you have added it to app but not sure why.

Sources/MQTTNIO/MQTTConnection.swift
```diff
@@ -75,7 +74,16 @@ final class MQTTConnection {
                         return channel.pipeline.addHandlers(handlers)
                     }
                 }
-                .connect(host: client.host, port: client.port)
+
+            let channelFuture: EventLoopFuture<Channel>
+
+            if client.port == 0 {
```
adam-fowler, line 80: Not a huge fan of the assume it is a unix domain path if port is zero, but yeah can't do this without a breaking change. When this gets merged we should add an issue to clean this up for the next major release

mosquitto/socket/readme.md
```diff
@@ -0,0 +1,7 @@
+A local mosquitto server using `mosquitto/config/mosquitto.conf` will create a `mosquitto.sock` socket in this directory.
```
adam-fowler, line 1: It is common practice to add a `.gitkeep` file in a folder you want to keep around.

adam-fowler (changes requested):
Code changes look ok, there are a few issues with testing though.
- Tests are not working when I run mosquitto via the shell script `scripts/mosquitto.sh`
- For Github CI to work I need to publish a docker file to kick off mosquitto. The Dockerfile for that `scripts/Dockerfile.ci` will need updated. Before then CI will continue to fail.

jpgrayson:
> Code changes look ok, there are a few issues with testing though.

Thank you for taking a look at this PR. I really appreciate mqtt-nio.

> Tests are not working when I run mosquitto via the shell script `scripts/mosquitto.sh`

I'm taking a look at this now. Sorry I missed it in the first go around.

The obstacle I'm running into is that it does not seem possible for a (linux) container to create a unix socket in a MacOS host's filesystem, which is what we would be asking the `mosquitto.sh` script to do. This makes intuitive sense because how could a linux unix socket exist in a MacOS host?

mosquitto reports:

```
14:53:04: Opening unix listen socket on path ./mosquitto/socket/mosquitto.sock.
14:53:04: Error binding unix socket: Not supported
```

> For Github CI to work I need to publish a docker file to kick off mosquitto. The Dockerfile for that `scripts/Dockerfile.ci` will need updated. Before then CI will continue to fail.

For the CI case, the mosquitto container should be able to share the socket with the Linux host. I will try to get this working.

-----

So there is one current test environment that will not work and several that do:

- [ ] `mosquitto.sh` on MacOS host
- [x] `mosquitto.sh` on Linux host
- [x] `docker-compose.yml` on MacOS host
- [x] `docker-compose.yml` on Linux host
- [x] native `mosquitto` on MacOS host
- [x] native `mosquitto` on Linux host 

Do you have a preferred approach for how to address the unix socket sharing problem with `mosquitto.sh` on MacOS?

One possibility would be to have `mosquitto.sh` run a native mosquitto instance instead of the current containerized approach. This may be palatable since `docker-compose.yml` provides a fully containerized test environment.

Sources/MQTTNIO/MQTTConnection.swift
jpgrayson: Agreed that this approach feels a bit icky.

We could have a new/alternative `init` that takes a `unixSockePath` instead of `host` and `port` parameters. I think this would still maintain compatibility and would be a better interface for initializing an `MQTTClient`, but since `MQTTClient.host` and `MQTTClient.port` are both public attributes, we could not fully hide this from the user.

jpgrayson:
I've submitted #147 to help solve CI with unix sockets. That PR modifies the Ci workflow to restart the mosquitto service container after mqtt-nio is checked-out. This allows the upstream eclipse-mosquitto image to be used. And for unix sockets, the container restart allows the mosquitto socket to be available at the expected `mqtt-nio/mosquitto/socket` path via a volume mount.

I'm hoping #147 will be approved and merged. I'll then rebase and update this PR so that CI works with the unix domain socket feature.

Sources/MQTTNIO/MQTTConnection.swift
jpgrayson: I've updated the PR with a change that adds a new convenience init for `MQTTClient` that takes a `unixSocketPath` parameter instead of `host` and `port`. This feels like an improvement and seems fine from a compatibility standpoint.

jpgrayson:
I think this PR is in pretty good shape at this point. Please let me know if you see any remaining obstacles. I appreciate the time you've taken to work with me on this, @adam-fowler.

adam-fowler:
Do you want to rebase your other changes in and I think we can merge this after that

adam-fowler:
So the homebrew version of mosquitto doesn't work out of the box
It appears you have to create a link to mosquitto manually as homebrew doesn't do it for you.

adam-fowler (approved):
I think this can go in now.
I'll add a comment to add the symbolic link for mosquitto. Unless you have some other method to ensure it is in the path

PR #147, opened 2023-11-20
ci: Use eclipse-mosquitto image instead of custom

jpgrayson:
A fundamental problem with service containers in GitHub Actions is that containers are started before the repository is checked out. This means that any configuration for the service that exists in the repo cannot simply be provided to the service container as volume mounts. [1]

The custom mqttnio-mosquitto image solved this problem by pre-populating the mosquitto configuration from the mqtt-nio repo such that the container could be run in CI without needing access to the mqtt-nio checkout.

But this leaves the problem of having to manually keep the custom mqttnio-mosquitto image updated with any changes in the repo.

A workaround for containers starting before checkouts is to add an explicit container restart step after the checkout step.

Using this workaround allows CI to use the upstream/canonical eclipse-mosquitto image instead of a custom image. The mosquitto configuration from the repo is thus always up to date and no manual image build/push is required.

[1]: https://github.com/orgs/community/discussions/42127

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,19 +55,29 @@ jobs:
     container:
       image: ${{ matrix.tag }}
     services:
-      # need to update docker image if you update the certificates
-      # run `build-docker-ci-image.sh` 
       mosquitto:
-        image: adamfowlerphoto/mqttnio-mosquitto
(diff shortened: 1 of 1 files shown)


adam-fowler:
This is great. Not having to build a custom docker image makes maintenance a lot easier.

PR #148, opened 2023-11-20
mosquitto.sh runs native mosquitto

jpgrayson:
I am posting this as a possible solution to the problem of how to run tests natively with a mosquitto service.

This changes `scripts/mosquitto.sh` to run a native mosquitto process instead of running mosquitto as a containerized service. This specifically allows mosquitto to create/bind a unix socket path that is available to natively running tests.

This is the best way forward I am seeing to enable tests running natively on MacOS to support unix sockets. But I also appreciate that it's a very different paradigm.

The container based approach works on Linux; i.e. a container can share a unix socket to a Linux host. So `mosquitto.sh` *could* be written to use containers on Linux and only use a native mosquitto process on MacOS. Let me know if that alternative seems right.

### DIFF
--- a/scripts/mosquitto.sh
+++ b/scripts/mosquitto.sh
@@ -1,83 +1,61 @@
 #!/bin/bash
 
-set -ex
+# Run mosquitto MQTT broker with configuration needed for the test suite.
+#
+# If mosquitto is installed locally, it will be run. Otherwise, fall back to
+# running a containerized version of mosquitto. The --container/-C option may be
+# supplied to force running a containerized mosquitto.
+#
+# N.B. On MacOS, a native mosquitto executable must be run to connect to
+# mosquitto using a unix domain socket.
 
-CONTAINER_ID="$(docker container ls | grep eclipse-mosquitto | awk {'print $1'})"
-COMMAND="$1"
-SCRIPTS=$(dirname $0)
-ROOT=$(dirname $SCRIPTS)
+set -eu -o pipefail
 
 usage()
 {
-    echo "Usage: mosquitto.sh [start] [stop] [status]"
+    echo "Usage: mosquitto.sh [--container|-C]"
     exit 2
 }
 
-run()
+run-installed-mosquitto()
 {
-    if [ -z "$CONTAINER_ID" ]; then
-        docker run \
-            -p 1883:1883 \
-            -p 1884:1884 \
-            -p 8883:8883 \
-            -p 8080:8080 \
-            -p 8081:8081 \
-            -v $(pwd)/$ROOT/mosquitto/config:/mosquitto/config \
-            -v $(pwd)/$ROOT/mosquitto/certs:/mosquitto/certs \
-            eclipse-mosquitto
-    else
-        echo "Mosquitto is already running"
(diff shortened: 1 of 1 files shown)


adam-fowler:
Is it not possible to have mosquitto bind a path internal to the docker container?

jpgrayson:
Mosquitto can bind a unix socket path internal to the docker container, but tests running in the host MacOS environment need access to that socket path.

Using a volume mount is the nominal way to share parts of the host and container filesystems, but a unix socket is not a regular file and cannot be shared between a Linux container and a MacOS host.

A unix socket can be shared between a container and a linux host.

AFAICT, the only way to enable tests running natively on MacOS to connect to mosquitto via a unix socket will involve running a native mosquitto process on the MacOS host.

This change modifies `mosquitto.sh` for this purpose.

As noted, we could structure this script to continue using the containerized mosquitto approach on Linux hosts where that will work with unix sockets.

adam-fowler:
This script was originally written on the understanding mosquitto was running inside Docker and might be running in the background. We can probably remove that and just support running in the foreground now.

jpgrayson:
I can make the script just run in the foreground. That would be simpler and would avoid monkeying with a pid file.

I'd kind of like to have the option on Linux of running either non-containerized mosquitto or in a container. Perhaps by testing for a `mosquitto` executable and falling back to running a container.

I'll update the PR with these changes. Let me know what you think. And thanks again for taking a look at these changes.

adam-fowler:
@jpgrayson this all sounds good.

scripts/mosquitto.sh
```diff
@@ -1,83 +1,58 @@
...
-        echo "Mosquitto is already running"
-    fi
+    docker run \
+        -p 1883:1883 \
+        -p 1884:1884 \
+        -p 8883:8883 \
+        -p 8080:8080 \
+        -p 8081:8081 \
+        -v ./mosquitto/config:/mosquitto/config \
+        -v ./mosquitto/certs:/mosquitto/certs \
```
adam-fowler, line 34: These folder mappings need a full path. It fails otherwise

scripts/mosquitto.sh
```diff
@@ -1,83 +1,58 @@
...
 fi
 
+cd "$(dirname "$(dirname "$0")")"
+
+if [[ $USE_CONTAINER -eq 1 ]]; then
+    run-containerized-mosquitto
+elif command -v mosquitto >/dev/null; then
+    run-installed-mosquitto
+else
+    echo "notice: mosquitto not installed; running eclipse-mosquitto container instead..."
```
adam-fowler, line 56: Maybe add how to install mosquitto

scripts/mosquitto.sh
jpgrayson: Interesting. The relative paths worked for me, but I'm using podman, not docker.

I changed this to absolute paths in the latest version of the change.

scripts/mosquitto.sh
jpgrayson: Done

adam-fowler (approved):

PR #149, opened 2023-11-22
Repair MQTTClient.v5.unsubscribe to use properties arg

jpgrayson:
A new/empty properties was being put into the unsubscribe packet instead of the user-provided properties argument.

Although there are not any nominal properties that would be used with unsubscribe packets, it is possible that user-defined properties could have vendor-specific meaning for certain MQTT brokers. [1]

[1]: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901182


adam-fowler:
Good catch

adam-fowler (approved):

PR #151, opened 2023-11-28
Update swiftformat

jpgrayson:
This updates the validation tooling to use `swiftformat` 0.52.10 (was 0.48.17).

I'm assuming there is value in updating tooling from time to time. If not, we can just close this PR.

This updated version of `swiftformat` does drive several formatting changes. The most common is to change, e.g.:

```swift
if let thing = thing {
    ...
}
```

to

```swift
if let thing {
    ...
```

which seems like a nice enough improvement.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -14,6 +14,6 @@ jobs:
     - name: Install Dependencies
       run: |
         brew install mint
-        mint install nicklockwood/swiftformat@0.48.17 --no-link
+        mint install nicklockwood/swiftformat@0.52.10 --no-link
     - name: run script
       run: ./scripts/validate.sh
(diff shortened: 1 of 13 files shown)


adam-fowler:
Is `if let thing` available in swift 5.7? I thought it came in 5.9. 
MQTTNIO supports the last three minor releases of Swift which include 5.7

jpgrayson:
Since `swiftformat` is configured for a particular minimum swift version, I was sort of trusting it to do the right thing in this regard.

The[ CI build/test with swift 5.7](https://github.com/swift-server-community/mqtt-nio/actions/runs/7019818778/job/19098139658?pr=151) also works. 

So it seems like this is okay.

adam-fowler:
Of course you're right. Thanks for all the changes.

adam-fowler (approved):

jpgrayson:
Thank you for taking the time to review these PR's. I very much appreciate your help!

Issue #156, opened 2024-09-18
test certificates are expired


jpgrayson:
**Describe the bug**

The test certificates found in `mosquitto/certs/` are expired. This causes the three SSL-related tests in the test suite to fail.

```
Test Case '-[MQTTNIOTests.MQTTNIOTests testSSLConnect]' started.
/Users/runner/work/mqtt-nio/mqtt-nio/.build/checkouts/swift-nio/Sources/NIOCore/EventLoopFuture.swift:1039: error: -[MQTTNIOTests.MQTTNIOTests testSSLConnect] : failed: caught error: "connectTimeout(NIOCore.TimeAmount(nanoseconds: 10000000000))"
Test Case '-[MQTTNIOTests.MQTTNIOTests testSSLConnect]' failed (10.402 seconds).
Test Case '-[MQTTNIOTests.MQTTNIOTests testSSLConnectFromP12]' started.
/Users/runner/work/mqtt-nio/mqtt-nio/.build/checkouts/swift-nio/Sources/NIOCore/EventLoopFuture.swift:1039: error: -[MQTTNIOTests.MQTTNIOTests testSSLConnectFromP12] : failed: caught error: "connectTimeout(NIOCore.TimeAmount(nanoseconds: 10000000000))"
Test Case '-[MQTTNIOTests.MQTTNIOTests testSSLConnectFromP12]' failed (10.129 seconds).
...
Received: Hello
Trust failed: “soto.codes” certificate is expired
Trust failed: “soto.codes” certificate is expired
Trust failed: “soto.codes” certificate is expired
```

**To Reproduce**

`swift test --filter testSSLConnect --filter testWebsocketAndSSLConnect`

**Expected behavior**

Tests should pass.

**Context (please complete the following information):**

This issue presents on MacOS and in the Linux CI environment.

**Additional context**

I believe @adam-fowler may be the only person that is able to generate the new certs. My naive attempt to run `scripts/generate-certs.sh` myself results in more abrupt SSL connection failures.

I would like to update the CI test matrix to include Swift 6.0, which is how I bumped into this problem.

adam-fowler:
Not sure why'd it be just me that can generate the certificates. I'll have a look into it and try and make sure anyone can generate the certificates.

jpgrayson:
My initial attempt to generate certs was using a version of openssl installed by homebrew.

[...]

PR #157, opened 2024-09-18
CI repair

jpgrayson:
- update testing certs (#156)
- update to `actions/checkout@v4` (resolves GitHub deprecation warning)
- add Swift 6.0 to test matrix

N.B. there remains a GitHub deprecation warning related to `codecov/codecov-action@v2`, which uses an out of date version of node. But to migrate to `codecov/codecov-action@v4` requires setting up a codecov token [as described here](https://docs.codecov.com/docs/adding-the-codecov-token), which requires a repo admin permissions.


adam-fowler:
I'm wondering if a better solution is to generate new certs in the GitHub action so they never expire for CI.

adam-fowler:
Also codecov wasn't even enabled on this repo, when it got moved to swift-server-community. You might as well up the version number on the action and I'll see if I can fix it

jpgrayson:
Added the change for codecov. You'll need to define a `CODECOV_TOKEN` secret as shown here:

https://docs.codecov.com/docs/adding-the-codecov-token

jpgrayson:
> I'm wondering if a better solution is to generate new certs in the GitHub action so they never expire for CI.

That's an idea. Given the trouble I had with needing to use MacOS' libreSSL versus homebrew's OpenSSL, it might be that the `generate-certs.sh` script needs some tweaks to be more compatible with the different openssl implementations out there.

jpgrayson:
bump @adam-fowler 

I think this PR is ready to go. The updated codecov seems to be working. The CI failure is due to codecov detecting -0.96% less coverage, but the process itself seems to be working.

I was not planning on trying to do the CI-time cert generation in this PR.

adam-fowler (approved):

adam-fowler:
Done. Thanks for this @jpgrayson

PR #159, opened 2024-10-08
Import FoundationEssentials instead of Foundation when available

Cyberbeni:
Removing Foundation dependency on non-Apple platforms allows us to significantly reduce the size of the build product when using the Swift Static Linux SDK.

Hello world docker image depending on mqtt-nio went from 169MB to 85MB.

Related PR in one of the dependencies: https://github.com/apple/swift-nio-transport-services/pull/209

### DIFF
--- a/Sources/MQTTNIO/AsyncAwaitSupport/MQTTClient+async.swift
+++ b/Sources/MQTTNIO/AsyncAwaitSupport/MQTTClient+async.swift
@@ -12,8 +12,8 @@
 //===----------------------------------------------------------------------===//
 
 #if canImport(FoundationEssentials)
-import FoundationEssentials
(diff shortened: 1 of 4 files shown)


adam-fowler:
Can you fix the validation issue here. Ideally you'd just run swiftformat but I haven't updated the version of swiftformat in a while so you might get a load of unrelated changes. 

Instead just reorder the FoundationEssentials and Dispatch imports

adam-fowler (approved):
I haven't had a chance to investigate how using Foundation vs FoundationEssentials affects a library but this looks good.

Cyberbeni:
This video gives a good explanation: https://youtu.be/wn6C_XEv1Mo

PR #189, opened 2026-01-05
Various small fixes

fpseverino:
- Remove `NIOConcurrencyHelpers` in favor of `Synchronization`
- Fix `docker-compose.yml` and devcontainer for Swift 6.1
- Remove unused DocC building scripts
- Update and fix `README.md` and `CONTRIBUTING.md`


Sources/MQTTNIO/MQTTInflight.swift
```diff
@@ -11,38 +11,36 @@
...
 //===----------------------------------------------------------------------===//
 
-import NIO
-import NIOConcurrencyHelpers
+import NIOCore
+import Synchronization
 
 /// Array of inflight packets. Used to resend packets when reconnecting to server
-struct MQTTInflight {
+struct MQTTInflight: Sendable, ~Copyable {
```
fpseverino: Without the `~Copyable` conformance it throws the error:
```
Stored property 'packets' of 'Copyable'-conforming struct 'MQTTInflight' has non-Copyable type 'Mutex<CircularBuffer<any MQTTPacket>>'

MQTTInflight.swift(18, 8): Consider adding '~Copyable' to struct 'MQTTInflight'
```
```swift
let packets: Mutex<CircularBuffer<any MQTTPacket>>
```

.devcontainer/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ version: "3.3"
 
 services:
   app:
-    image: swift:6.0
+    image: swift:6.1
```
adam-fowler: Might as well just use `swift:latest` to avoid having to update this all the time

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -520,7 +519,7 @@ public final actor MQTTConnection: Sendable {
 
     /// Resend PUBLISH and PUBREL messages
     func resendOnRestart() async throws {
-        let inflight = self.inflight.packets
+        let inflight = self.inflight.packets.withLock { $0 }
```
adam-fowler: Since inflight is only ever accessed from the MQTTConnection actor we can get rid of the lock.
We need to know what we want to do with these, as currently the inflight packets are lost. It might not even make sense to keep this around with the new structure of the package

fpseverino: Ok, I'll take a better look at them

adam-fowler: Remove the lock for the moment. We will need to look into the reconnecting story later.

adam-fowler (approved):

Issue #110, opened 2022-04-19
Change MQTTError into a struct


adam-fowler:
This requires a major version upgrade

adam-fowler:
Can't easily do this and keep the associated values. Perhaps we just wait for extensible enums

fpseverino:
> Perhaps we just wait for extensible enums

IIRC they should already be available in Swift 6.2.3, we could use `swift-tools-version:6.2.3`. Otherwise we wait for Swift 6.3

adam-fowler:
I think we should wait for 6.3.

PR #193, opened 2026-01-16
Cleanup MQTTError

adam-fowler:
Removed some unused errors, renamed `noConnection` to `connectionClosed`
I did have a look to see if I could change this to a `struct` but the different associated values make that very difficult. I think it probably best to wait for extensible enums


fpseverino:
See https://github.com/swift-server-community/mqtt-nio/issues/110#issuecomment-3759430759

fpseverino (approved):

PR #195, opened 2026-01-16
Adopt Swift 6 language mode and "NonisolatedNonsendingByDefault"

### LINKED ISSUE #188: v3: Adopt upcoming nonisolated(nonsending) by default
Add this to top of Package.swift

```swift
var defaultSwiftSettings: [SwiftSetting] =
    [
        .swiftLanguageMode(.v6),
    ]
#if compiler(>=6.2)
defaultSwiftSettings.append(
    .enableUpcomingFeature("NonisolatedNonsendingByDefault")
)
#endif
```

And make sure every target has their `swiftSettings` set `defaultSwiftSettings`

fpseverino:
- Bump `swift-tools-version` to 6.2
- Enable Swift 6 language mode and "NonisolatedNonsendingByDefault"
- Remove unused ping/pong code from WebSocketHandler that was causing warnings

There is a warning left in MQTTChannelHandler that right now I don't know how to solve and a few left in MQTTConnection that require a bit of restructuring (will do in a separate PR)

Resolves #188

### DIFF
--- a/.github/workflows/api-breakage.yml
+++ b/.github/workflows/api-breakage.yml
@@ -7,7 +7,7 @@ jobs:
   linux:
     runs-on: ubuntu-latest
     container:
-      image: swift:6.2
+      image: swift:latest
     steps:
     - name: Checkout
       uses: actions/checkout@v6
--- a/.github/workflows/api-breakage.yml
+++ b/.github/workflows/api-breakage.yml
@@ -7,7 +7,7 @@ jobs:
   linux:
     runs-on: ubuntu-latest
     container:
-      image: swift:6.1
+      image: swift:6.2
     steps:
     - name: Checkout
       uses: actions/checkout@v6
--- a/Package.swift
+++ b/Package.swift
@@ -4,8 +4,7 @@ import PackageDescription
 
 var defaultSwiftSettings: [SwiftSetting] =
     [
-        .swiftLanguageMode(.v6),
-        .enableUpcomingFeature("NonisolatedNonsendingByDefault"),
+        .enableUpcomingFeature("NonisolatedNonsendingByDefault")
     ]
 
 let package = Package(
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,12 @@
-// swift-tools-version:6.1
+// swift-tools-version:6.2
 
 import PackageDescription
+
+var defaultSwiftSettings: [SwiftSetting] =
+    [
+        .swiftLanguageMode(.v6),
+        .enableUpcomingFeature("NonisolatedNonsendingByDefault"),
+    ]
 
 let package = Package(
     name: "mqtt-nio",
@@ -24,9 +30,15 @@ let package = Package(
                 .product(name: "NIOWebSocket", package: "swift-nio"),
                 .product(name: "NIOSSL", package: "swift-nio-ssl", condition: .when(platforms: [.linux, .macOS])),
                 .product(name: "NIOTransportServices", package: "swift-nio-transport-services"),
-            ]
+            ],
+            swiftSettings: defaultSwiftSettings
         ),
-        .testTarget(name: "MQTTNIOTests", dependencies: ["MQTTNIO"]),
-    ],
-    swiftLanguageModes: [.v5]
+        .testTarget(
+            name: "MQTTNIOTests",
+            dependencies: [
+                .target(name: "MQTTNIO")
+            ],
+            swiftSettings: defaultSwiftSettings
+        ),
+    ]
 )
--- a/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
+++ b/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
@@ -413,6 +413,7 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
                 // otherwise reschedule task
                 if channelHandler.lastPingreqEventTime + channelHandler.pingreqTimeout <= .now() {
                     guard context.channel.isActive else { return }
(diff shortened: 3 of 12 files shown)


Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
fpseverino: This line throws the warning `Reference to captured var 'publish' in concurrently-executing code`.
How can we get the `publishMessage` out of the `sendMessage`'s `checkInbound` closure?

adam-fowler: Urghh that's a horrid bit of code. That's not an easy one to resolve

adam-fowler: Not totally sure what I'm doing there. It looks like I'm trying to deal with a PUBLISH being sent twice and resetting the PUBREC/PUBREL handshake. I don't think it even works. Perhaps comment out the mutable publish and the setting of publish if we receive a publish packet add a big comment to investigate and also add an issue here to check it out.

Package.swift
```diff
@@ -1,7 +1,13 @@
-// swift-tools-version:6.1
+// swift-tools-version:6.2
 
 import PackageDescription
 
+var defaultSwiftSettings: [SwiftSetting] =
+    [
+        .swiftLanguageMode(.v6),
```
adam-fowler: You don't need this if the tools version is set to 6.0 or above

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
```diff
@@ -170,41 +170,46 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
...
     /// If QoS is `.exactlyOnce` then send PUBREC, wait for PUBREL and then respond with PUBCOMP (in `respondToPubrel`)
-    private func respondToPublish(_ message: MQTTPublishPacket, context: ChannelHandlerContext) {
+    private func respondToPublish(_ message: MQTTPublishPacket, channelHandler: MQTTChannelHandler, context: ChannelHandlerContext) {
         switch message.publish.qos {
         case .atMostOnce:
             self.subscriptions.notify(message.publish)
 
         case .atLeastOnce:
+            let loopBoundChannelHandler = NIOLoopBound(channelHandler, eventLoop: self.eventLoop)
+            let loopBoundContext = NIOLoopBound(context, eventLoop: self.eventLoop)
```
adam-fowler: Instead of NIOLoopBound you can use `assumeIsolated()` on the EventLoopFuture returned by `context.channel.writeAndFlush(_:)`. It slightly cleaner

adam-fowler (commented):
There are a number of places where `assumeIsolated()` can be used instead of `NIOLoopBound`. It's easier to use and looks cleaner.

Also there are still two warnings about mutable types in `MQTTConnection.connect`, but I'd prefer you leave them as my NIOAsyncTestingChannel PR touches that function and I can fix in that PR to make the merge easier

fpseverino:
> Also there are still two warnings about mutable types in `MQTTConnection.connect`, but I'd prefer you leave them as my NIOAsyncTestingChannel PR touches that function and I can fix in that PR to make the merge easier

Yeah, they are due to the fact that we mutate the `configuration` and the `cleanSession` setting.

I'm studying clean sessions from the MQTT specification because I think we're doing something wrong at the moment (we could also move `cleanSession` and `identifier` inside `MQTTConnectionConfiguration`, WDYT?).

The mutation of `configuration` is because we have different settings for `keepAliveInterval` and `pingInterval`, what do you think if we merge the two into a single property (in a future PR)?

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
```diff
@@ -170,13 +170,14 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
     /// If QoS is `.atMostOnce` then no response is required
     /// If QoS is `.atLeastOnce` then send PUBACK
     /// If QoS is `.exactlyOnce` then send PUBREC, wait for PUBREL and then respond with PUBCOMP (in `respondToPubrel`)
-    private func respondToPublish(_ message: MQTTPublishPacket, context: ChannelHandlerContext) {
+    private func respondToPublish(_ message: MQTTPublishPacket, channelHandler: MQTTChannelHandler, context: ChannelHandlerContext) {
```
adam-fowler: You don't need the `channelHandler` function parameter anymore

.github/workflows/api-breakage.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   linux:
     runs-on: ubuntu-latest
     container:
-      image: swift:6.1
+      image: swift:6.2
```
adam-fowler:
```suggestion
      image: swift:latest
```

Sources/MQTTNIO/Subscriptions/MQTTConnection+subscribe.swift
```diff
@@ -54,7 +54,7 @@ extension MQTTConnection {
             throw MQTTError.cancelledTask
         }
         let packet = MQTTSubscribePacket(subscriptions: subscriptions, properties: properties, packetId: self.updatePacketId())
-        let subscriptionID: UInt32 = try await withCheckedThrowingContinuation { continuation in
+        let subscriptionID: UInt32 = try await withCheckedThrowingContinuation(isolation: self) { continuation in
```
adam-fowler: Is this necessary?

adam-fowler (commented):
Mostly looks good. Just a few minor things

adam-fowler (approved):

PR #197, opened 2026-01-16
Use NIOAsyncTestingChannel in tests

adam-fowler:
- Added `MQTTConnection.setupChannelAndConnect` to setup an arbitrary Channel 
- Split sending of connect packet into own function, so I can call it in tests. Also made the `close` function nonisolated so I can call it without await
- Added support for writing a `MQTTConnAckPacket`. A number of packets don't have a write or a read function because the client doesn't need them, but if we are to support faking a server we will need to fix these up. Perhaps one day that'll mean we can write a full server 🤷 
- Moved tests using a real server into IntegrationTests test target
- Added `MQTTConnectionTests` with one test to verify connect and disconnect works


Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -345,6 +351,64 @@ public final actor MQTTConnection: Sendable {
...
+
+    private static func _setupChannelAndConnect(
+        _ channel: Channel,
+        configuration: MQTTConnectionConfiguration,
+        cleanSession: Bool,
+        identifier: String,
+        logger: Logger
+    ) -> EventLoopFuture<MQTTConnection> {
+        do {
+            return channel.connect(to: try SocketAddress(ipAddress: "127.0.0.1", port: 6379)).flatMap {
```
```suggestion
            return channel.connect(to: try SocketAddress(ipAddress: "127.0.0.1", port: 1883)).flatMap {
```

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -345,6 +351,64 @@ public final actor MQTTConnection: Sendable {
...
+                        configuration: configuration,
+                        logger: logger
+                    )
+                    return MQTTConnection(
+                        channel: channel,
+                        channelHandler: handler,
+                        configuration: configuration,
+                        cleanSession: cleanSession,
+                        identifier: identifier,
+                        address: .hostname("127.0.0.1", port: 6379),
```
```suggestion
                        address: .hostname("127.0.0.1", port: 1883),
```

Tests/IntegrationTests/IntegrationTests.swift
```diff
@@ -0,0 +1,704 @@
...
+            #if os(Linux)
+            MultiThreadedEventLoopGroup.singleton
+            #else
+            // Return TS Eventloop for non-Linux builds, as we use TS TLS
+            NIOTSEventLoopGroup.singleton
+            #endif
+        }
+
+        let logger: Logger = {
+            var logger = Logger(label: "MQTTNIOTests")
```
fpseverino:
```suggestion
            var logger = Logger(label: "IntegrationTLSTests")
```

This makes it easier to read the test logs

Tests/IntegrationTests/IntegrationTests.swift
```diff
@@ -0,0 +1,704 @@
...
+                    try await connection.publish(to: "overlapping/home/kitchen/temperature", payload: ByteBuffer(string: "test"), qos: .atLeastOnce)
+                }
+
+                try await group.waitForAll()
+            }
+        }
+    }
+
+    let logger: Logger = {
+        var logger = Logger(label: "MQTTNIOTests")
```
fpseverino:
```suggestion
        var logger = Logger(label: "IntegrationTests")
```

As above, we should also do this for "IntegrationV5Tests"

fpseverino (approved):
Nothing blocking, just a few nits

Sources/MQTTNIO/Connection/MQTTConnection.swift
adam-fowler: I didn't copy that from the valkey repo honest

PR #201, opened 2026-01-17
Cancellation

### LINKED ISSUE #187: v3: Cancellation
We should be able to cancel the sending of a message (or at least the waiting for a response). The way this can be done is by creating a unique id for each message that is sent.

Wrap the contents of `sendMessage` with `withTaskCancellationHandler`. The `onCancel` closure should call a `cancel` function with the unique id on the ChannelHandler which will call a similar function on its stateMachine. Inside the stateMachine we'll find the task with the id and then fail its future.

fpseverino:
Resolves #187

### DIFF
--- a/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
+++ b/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
@@ -39,17 +39,16 @@ extension MQTTChannelHandler {
             let context: Context
             var tasks: [MQTTTask]
 
-            func cancel(requestID: Int) -> (cancel: [MQTTTask], connectionClosedDueToCancellation: [MQTTTask]) {
-                var withRequestID = [MQTTTask]()
-                var withoutRequestID = [MQTTTask]()
-                for task in tasks {
+            mutating func cancel(requestID: Int) -> [MQTTTask] {
+                var cancelledTasks: [MQTTTask] = []
+                for task in self.tasks {
                     if task.requestID == requestID {
-                        withRequestID.append(task)
-                    } else {
-                        withoutRequestID.append(task)
+                        cancelledTasks.append(task)
+                        self.tasks.removeAll { $0 === task }
+                        // TODO: bad performance, will update after #202 is merged
                     }
                 }
-                return (withRequestID, withoutRequestID)
+                return cancelledTasks
             }
         }
 
@@ -197,7 +196,7 @@ extension MQTTChannelHandler {
 
         @usableFromInline
         enum CancelAction {
-            case failTasksAndClose(Context, cancel: [MQTTTask], closeConnectionDueToCancel: [MQTTTask])
+            case failTasks([MQTTTask])
             case doNothing
         }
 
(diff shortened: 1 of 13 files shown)


Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
```diff
@@ -38,6 +38,19 @@ extension MQTTChannelHandler {
         struct InitializedState {
             let context: Context
             var tasks: [MQTTTask]
+
+            func cancel(requestID: Int) -> (cancel: [MQTTTask], connectionClosedDueToCancellation: [MQTTTask]) {
```
adam-fowler: It might be a good idea to wait until #202 is merged as this has a more optimal way of removing tasks

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
```diff
@@ -182,6 +195,37 @@ extension MQTTChannelHandler {
             }
         }
 
+        @usableFromInline
+        enum CancelAction {
+            case failTasksAndClose(Context, cancel: [MQTTTask], closeConnectionDueToCancel: [MQTTTask])
```
adam-fowler: I don't think we need to close the connection on cancellation. We do in valkey-swift as cancellation leaves the connection in an indeterminate state eg you could have kicked off a blocking command and the connection cannot be used until it unblocks. I don't think MQTT has anything similar.

Sources/MQTTNIO/MQTTError.swift
```diff
@@ -66,6 +66,8 @@ public enum MQTTError: Error, Sendable {
     case versionMismatch(expected: MQTTConnectionConfiguration.Version, actual: MQTTConnectionConfiguration.Version)
     /// The Task was cancelled
     case cancelledTask
+    /// Connection closed because another task was cancelled.
+    case connectionClosedDueToCancellation
```
adam-fowler: As I said before not sure we need to close the connection on cancellation

Tests/MQTTNIOTests/MQTTConnectionTests.swift
```diff
@@ -491,6 +491,47 @@ struct MQTTConnectionTests {
             try await channel.writeInboundPacket(finalAuth, version: .v5_0)
         }
     }
+
+    @Test("Connection Close due to Cancellation")
+    func connectionCloseDueToCancellation() async throws {
```
adam-fowler: For this test can you use `withTestMQTTServer`. It ensures both CONNECT and DISCONNECT are sent. You can use an AsyncStream to communicate between the two closures eg
```swift
let (stream, cont) = AsyncStream.makeStream(of: Void.self)
try await withTestMQTTServer(logger: logger) { connection in
    try await withThrowingTaskGroup { group in
        group.addTask {
            await #expect(throws: MQTTError.cancelledTask) {
                try await connection.publish(to: "foo", payload: ByteBuffer(), qos: .exactlyOnce)
            }
        }
        await stream.first { _ in true }
        group.cancelAll()
    }
} server: { _ in
    _ = try await channel.waitForOutboundWrite(as: ByteBuffer.self)
    cont.yield()
}
```

Tests/IntegrationTests/IntegrationTests.swift
```diff
@@ -650,6 +650,50 @@ struct IntegrationTests {
         }
     }
 
+    @Test("Cancellation")
+    func cancellation() async throws {
```
adam-fowler, line 654: I'm not sure the cancellation integration tests provide much value. All they are doing is ensuring the subscription AsyncStream ends on cancellation. It doesn't actually test the cancellation code you wrote above

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
```diff
@@ -38,6 +38,19 @@ extension MQTTChannelHandler {
         struct InitializedState {
             let context: Context
             var tasks: [MQTTTask]
+
+            mutating func cancel(requestID: Int) -> [MQTTTask] {
```
adam-fowler: You should only ever have one cancelled task

adam-fowler (commented):
Cancellation should only ever return one task

adam-fowler (approved):

PR #202, opened 2026-01-17
Fix handling of duplicate PUBLISH packets in QoS 2

### LINKED ISSUE #196: v3: Investigate what to do if we receive PUBLISH while waiting for PUBREL
In MQTTChannelHandler, in the `respondToPublish` method, in the `exactlyOnce` case (QoS 2), if we received a PUBLISH message while waiting for a PUBREL from broker, the behavior was to replace data to be published and retry PUBREC.

Investigate if it works and how to do it without warnings in Swift 6 language mode.

See https://github.com/swift-server-community/mqtt-nio/pull/195#discussion_r2698309007

adam-fowler:
If the server doesn't receive a PUBREC in time, it will send another PUBLISH tagged as a duplicate and will expect the PUBREC,PUBREL,PUBCOMP handshake to be repeated. Previously MQTTNIO was treating this as a complete new PUBLISH and the old PUBLISH was never completed. If this was repeated the task array could grow indefinitely.

To fix this now if a PUBLISH is tagged as a duplicate we check to see if there is already a PUBLISH with the same packet id being processed if so we pass the new PUBLISH to the associated MQTTTask. The task then will see it has received a PUBLISH when it was expecting a PUBREL. It will quit that task (by throwing an error) and restart the PUBREC,PUBREL,PUBCOMP handshake.

I also improved the performance of removing a task from the task array by copying the last task into the slot that was being removed, instead of calling `removeAll`

### DIFF
--- a/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
+++ b/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
@@ -121,16 +121,18 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
         let buffer = self.unwrapInboundIn(data)
         do {
             try self.decoder.process(buffer: buffer) { message in
+                self.logger.trace(
+                    "MQTT In",
+                    metadata: ["mqtt_message": .stringConvertible(message), "mqtt_packet_id": .stringConvertible(message.packetId)]
+                )
                 switch self.stateMachine.receivedPacket(message) {
                 case .respondAndReturn:
                     let publishMessage = message as! MQTTPublishPacket
                     // publish logging includes topic name
                     self.logger.trace(
-                        "MQTT In",
+                        "MQTT Publish In",
                         metadata: [
-                            "mqtt_message": .stringConvertible(publishMessage),
-                            "mqtt_packet_id": .stringConvertible(publishMessage.packetId),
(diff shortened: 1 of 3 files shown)


Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
fpseverino:
```suggestion
```

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
```diff
@@ -196,10 +195,11 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
             //var publish = message.publish
```
fpseverino: These comments can be removed

Tests/MQTTNIOTests/MQTTConnectionTests.swift
```diff
@@ -343,6 +343,64 @@ struct MQTTConnectionTests {
         }
     }
 
+    @Test
+    func testSubscribeAndPublishQoS2MissedPUBREC() async throws {
+        var logger = Logger(label: "testSubscribeAndPublishQoS2")
```
fpseverino:
```suggestion
        var logger = Logger(label: "testSubscribeAndPublishQoS2MissedPUBREC")
```

fpseverino (approved):

PR #207, opened 2026-01-20
Improve TLS connection configuration

fpseverino:
This PR updates the TLS configuration within `MQTTConnectionConfiguration` so that it is a combination of that of `valkey-swift` (with two options, `.disable` and `.enable(TLSConfig, tlsServerName)`) and the old one from MQTTNIO (allowing native TLS on both NIOSSL and Network/Security frameworks).

This may need to change again in the future, as currently the `NIOSSLContext` is created for each connection during the connection phase, and when we have the `MQTTClient` with the connection pool, we may want to cache the `NIOSSLContext`.

Also, I've combined the `_tlsConfiguration` and `getTLSConfiguration` testing utilities.

### DIFF
--- a/Sources/MQTTNIO/Connection/MQTTConnection.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnection.swift
@@ -438,27 +438,36 @@ public final actor MQTTConnection: Sendable {
         host: String,
         logger: Logger
     ) throws -> NIOClientTCPBootstrap {
-        var bootstrap: NIOClientTCPBootstrap
-        let serverName = configuration.sniServerName ?? host
+        var serverName: String {
+            if case .enable(_, let sniServerName) = configuration.tls.base, let sniServerName {
+                sniServerName
+            } else {
+                host
+            }
+        }
+
+        let bootstrap: NIOClientTCPBootstrap
         #if canImport(Network)
         // if eventLoop is compatible with NIOTransportServices create a NIOTSConnectionBootstrap
         if let tsBootstrap = NIOTSConnectionBootstrap(validatingGroup: eventLoopGroup) {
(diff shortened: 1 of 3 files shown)


Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
```diff
@@ -68,20 +67,47 @@ public struct MQTTConnectionConfiguration: Sendable {
...
+            case niossl(TLSConfiguration)
+            #endif
+            #if canImport(Network)
+            /// NIO Transport Services TLS configuration.
+            case ts(TSTLSConfiguration)
+            #endif
+        }
+        enum Base {
+            case disable
+            case enable(Configuration, String?)
```
adam-fowler: Can we have a label for the `serverName`

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -438,27 +438,36 @@ public final actor MQTTConnection: Sendable {
...
-            case .niossl:
-                throw MQTTError.wrongTLSConfig
-            #endif
-            default:
+            if case .enable(let tlsConfigType, _) = configuration.tls.base {
+                switch tlsConfigType {
+                case .ts(let tsConfig):
+                    options = try tsConfig.getNWProtocolTLSOptions(logger: logger)
+                #if os(macOS) || os(Linux) || os(Android)
+                case .niossl:
```
adam-fowler, line 460: I'm wondering if we should allow NIOSSL TLS on transport services event loops. It works, it is just preferable you use the TLS already available with the network framework. I'll add an issue for that

adam-fowler (commented):
One minor thing

adam-fowler (approved):

PR #208, opened 2026-01-20
Improve authentication configuration

fpseverino:
Depends on #207 

This PR moves the two `userName` and `password` properties of `MQTTConnectionConfiguration` inside a `Authentication` struct, ensuring the user can send a user name without a password, but not a password without a user name, as per the spec


adam-fowler:
You've included the TLS changes here as well. I'll review once you've rebased with main

fpseverino:
> I'll review once you've rebased with main

@adam-fowler done

adam-fowler (approved):

PR #210, opened 2026-01-24
Report whether a session is present in `withConnection` closure

### LINKED ISSUE #206: Report whether session is present in withConnection closure
When you receive a CONNACK it reports whether there is already a session for that connection. This should be returned to the user eg

```
MQTTConnection.withConnection { connection, sessionPresent in
    ...
}
```

fpseverino:
Resolves #206

### DIFF
--- a/Sources/MQTTNIO/Connection/MQTTConnection.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnection.swift
@@ -106,6 +106,38 @@ public final actor MQTTConnection: Sendable {
         return try await operation(connection, sessionPresent)
     }
 
+    /// Connect to MQTT server and run operations using the connection and then close it.
+    ///
+    /// - Parameters:
+    ///   - address: Internet address of the MQTT server.
+    ///   - configuration: Configuration of the MQTT connection.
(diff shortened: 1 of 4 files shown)


Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -83,6 +83,7 @@ public final actor MQTTConnection: Sendable {
     ///   - eventLoop: EventLoop to run the connection on.
     ///   - logger: Logger to use for the connection.
     ///   - operation: Closure handling the MQTT connection.
+    ///     The closure receives the ``MQTTConnection`` and a `Bool` indicating if there was a previous session present.
     /// - Returns: Value returned from the operation closure.
     public static func withConnection<Value>(
```
adam-fowler, line 88: A lot of the time people won't care about the `sessionPresent` flag. Perhaps we include two withConnection functions one that has a closure that includes both parameters and the other which only includes a connection parameter

adam-fowler (commented):
All looks good, apart from one minor suggestion

adam-fowler (approved):

PR #212, opened 2026-01-27
Don't send packets exceeding `maxPacketSize`

### LINKED ISSUE #48: The Client MUST NOT send packets exceeding Maximum Packet Size to the Server
From spec: The Client MUST NOT send packets exceeding Maximum Packet Size to the Server

fpseverino:
Resolves #48


Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
```diff
@@ -106,14 +109,17 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
 
     func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
         let message = unwrapOutboundIn(data)
-        self.logger.trace("MQTT Out", metadata: ["mqtt_message": .string("\(message)"), "mqtt_packet_id": .string("\(message.packetId)")])
         var bb = context.channel.allocator.buffer(capacity: 0)
         do {
             try message.write(version: self.configuration.version, to: &bb)
+            guard bb.readableBytes <= (self.maxPacketSize ?? UInt32.max) else {
```
adam-fowler: Why don't you make maxPacketSize non optional and default it to `.max`

adam-fowler:
There was a reason why I never implemented this. I can't for the life of me remember why though. I'll have to have a think about this before approving.

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
fpseverino: Yes, I did that at first, but then I made it optional because I thought that if the server doesn't set it, _conceptually_ we don't have a `maxPacketSize`. However, I can make it non-optional and default to `UInt32.max`

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
```diff
@@ -364,7 +374,10 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
 
         switch self.stateMachine.sendPacket(task) {
         case .sendPacket(let context):
-            _ = context.channel.writeAndFlush(message)
+            context.channel.writeAndFlush(message).assumeIsolated().whenFailure { error in
+                task.fail(error)
+                self.stateMachine.removeTask(task)
```
adam-fowler: Could you not call `cancelTask` here, it does the same and then returns whether you should fail the promise. There is a chance the task has already been removed through another error.

fpseverino: You mean something like this? https://github.com/swift-server-community/mqtt-nio/pull/212/commits/4edcfccc058b9937388ff05400b01995c36840bb

adam-fowler (approved):

PR #217, opened 2026-02-02
V2: Add support for Android

adam-fowler:


.github/workflows/ci.yml
```diff
@@ -89,15 +99,22 @@ jobs:
...
-      uses: codecov/codecov-action@v4
+        swift test --enable-code-coverage
+    - name: Upload coverage data
+      uses: vapor/swift-codecov-action@v0.3
       with:
-        files: info.lcov
-        token: ${{ secrets.CODECOV_TOKEN }}
+        codecov_token: ${{ secrets.CODECOV_TOKEN }}
+  android:
+    name: Android
```
fpseverino:
```suggestion
        codecov_token: ${{ secrets.CODECOV_TOKEN }}

  android-build:
    name: Android Build
```

mosquitto/config/mosquitto.conf
```diff
@@ -1,6 +1,6 @@
 # Setup
-per_listener_settings true
-allow_zero_length_clientid true
+#per_listener_settings true
+#allow_zero_length_clientid true
```
fpseverino, line 3: Since we are keeping v2.0.22 of mosquitto, do we need/have to remove these settings?

Sources/MQTTNIO/MQTTClient.swift
```diff
@@ -21,7 +21,7 @@ import NIOConcurrencyHelpers
 import Network
 import NIOTransportServices
 #endif
-#if os(macOS) || os(Linux)
+#if !os(iOS)
```
fpseverino: Wouldn't checking only that it's *not* iOS cause problems on tvOS and watchOS? Or does the `os(iOS)` flag also cover these other operating systems?

We could just add `|| os(Android)`, which would also be the same logic used in the package manifest
```swift
.product(name: "NIOSSL", package: "swift-nio-ssl", condition: .when(platforms: [.linux, .macOS, .android])),
```

fpseverino (commented):
Just a couple of nits and some concern about OS checks

Sources/MQTTNIO/MQTTClient.swift
adam-fowler: This doesn't run on watchOS. The idea was if a new platform appears we don't need to add another platform to the list. But I guess if another platform appears it is better we actually respond to it

fpseverino (approved):

PR #219, opened 2026-02-05
V3: Android support

adam-fowler:


.github/workflows/ci.yml
```diff
@@ -17,7 +17,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  macos-unit:
+  macos-build:
+    name: macOS Build
```
fpseverino: My reasoning when I changed the job names was to call them `platform-unit` for the platforms where we run tests and `platform-build` for those where we simply check that it compiles, but if you want to change them, I'm not precious about it

fpseverino (approved):
Just a reminder that if job names are changed, the required status checks have to be updated

.github/workflows/ci.yml
adam-fowler: yeah I guess that makes sense. I'll change it

Issue #215, opened 2026-01-28
v3: MQTTSession


adam-fowler:
It is possible for an MQTT session to last between connections. Because of this it is worthwhile making the session a first class object. `MQTTSession` will store a client id, inflight publish packets and any subscriptions that have been made. It should be possible to call `subscribe` on `MQTTSession` and the subscription will be valid across multiple MQTT connections.

`MQTTConnection.withConnection` will get a new optional parameter `session`. This session will be stored in the connection. If no session is supplied then a new session is created. If you supply a session then the connection will be made asking for a session from the server (cleanSession/cleanStart = false). If the server's CONNACK indicates there is no session, then the MQTTSession contents are cleared and any subscription streams are ended. Otherwise the connection will continue with the session data it has got and any inflight packets will be resent.

Because you can call `subscribe` on `MQTTSession` outside of the `MQTTConnection.withConnection` this function cannot return as soon as its closure is done. It can only exit if all subscriptions are ended, an error is thrown or the connection is closed by the server.

```swift
let session = MQTTSession(name: "TestSession")
try await withThrowingTaskGroup { group in
    group.addTask {
        // this subscribe won't activate until a connection is made 
        try await session.subscribe(to: [.init(topicFilter: "testMQTTPublishRetain", qos: .atLeastOnce)]) { sub in
            for try await publish in sub {
                ...
            }
        }
    group.addTask {
        // does not exit until either all subscriptions are ended or there is an error
        try await MQTTConnection.withConnection(
            address: .hostname(Self.hostname, port: 8080),
            session: session
        ) { connection in
        }
    }
}
```

PR #220, opened 2026-02-06
Introduce `MQTTSession` for inflight messages

### LINKED ISSUE #204: v3: Review MQTTInflight and bring back tests
See https://github.com/swift-server-community/mqtt-nio/pull/189#discussion_r2663857209

fpseverino:
A first step towards #215, resolves #204

- Adds `MQTTSession` to manage session state across connections
    - right now only for inflight messages, moves logic from `MQTTInflight` to `MQTTSession`
- Splits `withConnection` into two different methods:
    - one takes a non-optional `MQTTSession`, sets the identifier as the one in the session and `cleanSession` to false
    - another one takes only an `identifier` and sets `cleanSession` to true, for users that don't need to keep session state but want to set the Client Identifier
- Removes `identifier` and `cleanSession` from `MQTTConnection`
    - if a v5 server sets the Assigned Client Identifier property in CONNACK, the identifier in `MQTTSession` gets updated
    - now the `identifier` is no longer accessible from the `MQTTConnection`, if the user wishes to read the Assigned Client Identifier, they must read it from `MQTTSession`
- Previously, if the user set the Session Expiry Interval when connecting, `cleanSession` was set automatically to `false`; this is no longer the case
- Adds tests for inflight messages and updates existing session tests


Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -45,41 +43,35 @@ public final actor MQTTConnection: Sendable {
     let channelHandler: MQTTChannelHandler
     let configuration: MQTTConnectionConfiguration
     let globalPacketId = Atomic<UInt16>(1)
-    /// Inflight messages
-    private var inflight: MQTTInflight
     let isClosed: Atomic<Bool>
 
     var connectionParameters = ConnectionParameters()
-    let cleanSession: Bool
+    let session: MQTTSession?
```
adam-fowler: Given the session is going to store all the subscriptions, I think an MQTTConnection should always have a session. If no session is supplied in `withConnection` we create a new one

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -130,7 +119,66 @@ public final actor MQTTConnection: Sendable {
...
+    /// - Returns: Value returned from the operation closure.
+    public static func withConnection<Value>(
+        address: MQTTServerAddress,
+        configuration: MQTTConnectionConfiguration = .init(),
+        session: MQTTSession,
+        eventLoop: any EventLoop = MultiThreadedEventLoopGroup.singleton.any(),
+        logger: Logger,
+        operation: (MQTTConnection, Bool) async throws -> sending Value
+    ) async throws -> sending Value {
+        let (connection, sessionPresent) = try await self.connect(
```
adam-fowler, line 151: Should we add a check to ensure we are not creating multiple connections to one session

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -805,12 +845,12 @@ public final actor MQTTConnection: Sendable {
             return nil
         }
 
-        self.inflight.add(packet: packet)
+        self.session?.addInflight(packet: packet)
```
adam-fowler: This requires a lock for every PUBLISH, PUBREL packet which isn't ideal. Perhaps we store a local copy of inflight packets and copy them across to the session on connection close. We only need them to persist in the session between connections

fpseverino: I did it in [this commit](https://github.com/swift-server-community/mqtt-nio/pull/220/commits/189a830eac75fe2886acb6bd03612688587066ba). The unfortunate thing is that copying the inflight messages to the `MQTTSession`, unlike sending `DISCONNECT`, cannot be done either in the `nonisolated` `connection.close()` method or in `defer` blocks, because we have to explicitly `await` for the inflight messages to be copied before exiting the `withConnection` method.
If we open a new connection with the same `MQTTSession` before the local copy of the old connection's inflights is copied back into the session, the new connection copies the outdated inflights from the session. At first, the tests failed for this reason.

Sources/MQTTNIO/Connection/MQTTConnection.swift
fpseverino: Done in https://github.com/swift-server-community/mqtt-nio/pull/220/commits/dec27e68fd72168a8a2b72eba49550e6bb1e7901

Sources/MQTTNIO/Connection/MQTTConnection.swift
adam-fowler: Perhaps we have an internal version of `close` that doesn't use execute, so we can guarantee it has fully closed before exiting `withConnection`. eg

```swift
private func _close() throws {
    try self.sendDisconnect()
    self.channel.close(mode: .all, promise: nil)
    self.saveInflightToSession()
}

public nonisolated func close() {
    ...
    self.channel.eventLoop.execute {
    self.assumeIsolated {
        try? $0._close()
    }
}
```

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -45,41 +43,38 @@ public final actor MQTTConnection: Sendable {
     let channelHandler: MQTTChannelHandler
     let configuration: MQTTConnectionConfiguration
     let globalPacketId = Atomic<UInt16>(1)
-    /// Inflight messages
-    private var inflight: MQTTInflight
+    /// Local copy of inflight messages
+    private var inflight: MQTTInflight?
```
adam-fowler: Any reason this is optional? It looks like if it is set to nil at initialisation inflight messages are never saved

Sources/MQTTNIO/Connection/MQTTConnection.swift
fpseverino: Done it in https://github.com/swift-server-community/mqtt-nio/pull/220/commits/abeb543b1fc663905380c8dce618e66816616591

Sources/MQTTNIO/Connection/MQTTConnection.swift
fpseverino: It was because `session` was also optional, I made it non-optional in https://github.com/swift-server-community/mqtt-nio/pull/220/commits/abeb543b1fc663905380c8dce618e66816616591

Sources/MQTTNIO/Connection/MQTTConnection.swift
fpseverino: We would still have to `await` the internal method, so no `defer` blocks, plus, if we combine `saveInflightToSession()` and `close()`, when we call `try? await connection._close()` inside the `catch` block, we could send `DISCONNECT` and close the channel a second time

adam-fowler (approved):
Looks good

PR #221, opened 2026-02-13
Enable various Swift settings

fpseverino:
New upcoming features enabled:
- ExistentialAny
- InternalImportsByDefault
- MemberImportVisibility
- InferIsolatedConformances
- ImmutableWeakCaptures

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,11 +4,22 @@ import PackageDescription
 
 var defaultSwiftSettings: [SwiftSetting] =
     [
+        // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0335-existential-any.md
         .enableUpcomingFeature("ExistentialAny"),
+
+        // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md
         .enableUpcomingFeature("InternalImportsByDefault"),
+
+        // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md
         .enableUpcomingFeature("MemberImportVisibility"),
+
+        // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0470-isolated-conformances.md
         .enableUpcomingFeature("InferIsolatedConformances"),
+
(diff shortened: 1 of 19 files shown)


Package.swift
```diff
@@ -4,7 +4,12 @@ import PackageDescription
 
 var defaultSwiftSettings: [SwiftSetting] =
     [
-        .enableUpcomingFeature("NonisolatedNonsendingByDefault")
+        .enableUpcomingFeature("ExistentialAny"),
```
adam-fowler, line 8: I think it is always nice to include a comment with a link to the related evolution document for each of these.

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
```diff
@@ -11,11 +11,11 @@
 //
 //===----------------------------------------------------------------------===//
 
-import NIOCore
-import NIOHTTP1
+public import NIOCore
+public import NIOHTTP1
```
adam-fowler, line 15: Perhaps we should be using the swift http headers in our public interfaces instead of NIOs. We can do this in another PR

Tests/IntegrationTests/IntegrationTests.swift
```diff
@@ -501,47 +501,6 @@ struct IntegrationTests {
         }
     }
 
-    @Test("WebSocket Initial Request")
```
adam-fowler, line 504: Why are we deleting this?

fpseverino: I moved it to the other target, together with other `NIOEmbedded` tests

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
fpseverino: You mean [swift-http-types](https://github.com/apple/swift-http-types)?

adam-fowler: Yeah that's the one.

adam-fowler (approved):

PR #222, opened 2026-02-23
Add `MQTTSession.subscribe`

fpseverino:


Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -158,7 +158,35 @@ public final actor MQTTConnection: Sendable {
             logger: logger
         )
         do {
-            let result = try await operation(connection, sessionPresent)
+            let result = try await withThrowingTaskGroup { group in
+                group.addTask {
+                    for await queuedSubscription in session.subscriptionsQueue {
```
adam-fowler: It might be better to combine the subscribe and unsubscribe queues. And use an enum with associated value as the element
```swift
enum SubscriptionTask {
    case subscribe(MQTTSubscribePacket)
    case unsubscribe(MQTTUnsubscribePacket)
}
```

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -186,7 +214,7 @@ public final actor MQTTConnection: Sendable {
         eventLoop: any EventLoop = MultiThreadedEventLoopGroup.singleton.any(),
         logger: Logger,
         operation: (MQTTConnection) async throws -> sending Value
```
adam-fowler: Given we are using non isolated(non sending) we can get rid of the sending here

adam-fowler (commented):
Haven't had a proper look yet, but a couple of initial comments

fpseverino:
Thank you for your valuable feedback @adam-fowler. The code currently has a problem, namely that after the first connection uses the session, the `AsyncStream`s are terminated and cannot be reused by subsequent connections. I am therefore working on a new type that keeps an internal buffer and generates new `AsyncStream`s for each connection.

adam-fowler:
> Thank you for your valuable feedback @adam-fowler. The code currently has a problem, namely that after the first connection uses the session, the `AsyncStream`s are terminated and cannot be reused by subsequent connections. I am therefore working on a new type that keeps an internal buffer and generates new `AsyncStream`s for each connection.

The connection closin shouldn't end the AsyncStreams setup by MQTTSession

fpseverino:
> The connection closin shouldn't end the AsyncStreams setup by MQTTSession

Cancelling the `TaskGroup` task where they are run terminates them

adam-fowler:
> > The connection closin shouldn't end the AsyncStreams setup by MQTTSession
> 
> Cancelling the `TaskGroup` task where they are run terminates them

Cancelling the `TaskGroup` task where they are iterated will terminate them, but if you are iterating an AsyncStream from the MQTTSession that shouldn't cancel when you close the connection. Let me look at the code closer and I'll get back to you.

adam-fowler:
So I wrote a small test and the session AsyncStream isn't ended on connection close (unless the subscribe is done from the connection).
```swift
await withThrowingTaskGroup { group in
    group.addTask {
        try await session.subscribe(to: [.init(topicFilter: "subscribeWithSession", qos: .atMostOnce)]) { subscription in
            var iterator = subscription.makeAsyncIterator()
            try #expect(await (iterator.next()?.payload).map { String(buffer: $0) } == "test")
            try #expect(await (iterator.next()?.payload).map { String(buffer: $0) } == "test2")
        }
    }

    group.addTask {
        try await MQTTConnection.withConnection(...) { connection, sessionPresent in
            try await connection.publish(to: "subscribeWithSession", payload: ByteBuffer(string: "test"), qos: .atMostOnce)
        }
        try await MQTTConnection.withConnection(...) { connection, sessionPresent in
            #expect(sessionPresent)
            try await connection.publish(to: "subscribeWithSession", payload: ByteBuffer(string: "test2"), qos: .atMostOnce)
        }
    }
}
```

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -158,7 +158,35 @@ public final actor MQTTConnection: Sendable {
             logger: logger
         )
         do {
-            let result = try await operation(connection, sessionPresent)
+            let result = try await withThrowingTaskGroup { group in
+                group.addTask {
+                    for await task in session.subscriptionsQueue {
```
adam-fowler: If you move this block of code to a member function of MQTTConnection, you can ditch the `connection.channel.eventLoop.execute` lower down

Sources/MQTTNIO/Session/MQTTSession+subscribe.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+// See LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension MQTTSession {
+    @inlinable
+    public nonisolated func subscribe<Value>(
```
adam-fowler: MQTTSession isn't an actor so we don't need `nonisolated` here

Sources/MQTTNIO/Session/MQTTSession.swift
```diff
@@ -25,6 +26,12 @@ public final class MQTTSession: Sendable {
     /// Inflight messages
     let inflightPackets: Mutex<MQTTInflight>
 
+    let subscriptions: Mutex<MQTTSubscriptions>
```
adam-fowler, line 29: Can we add a single mutex around all the three parameters (clientID, inflight and subscriptions). It would be very rare they are accessed simultaneously

Sources/MQTTNIO/Session/MQTTSession.swift
```diff
@@ -69,3 +80,21 @@ extension MQTTSession {
...
+        let id: UInt32
+        let continuation: MQTTSubscription.Continuation
+        let subscriptions: [MQTTSubscribeInfoV5]
+        let properties: MQTTProperties
+    }
+
+    @usableFromInline
+    enum SessionSubscriptionTask: Sendable {
+        case subscribe(QueuedSubscription)
+        case unsubscribe(UInt32, MQTTProperties)
```
adam-fowler: Given you have a struct for subscribe can we have one for unsubscribe as well. Can we have some comments as well

Tests/MQTTNIOTests/MQTTSubscriptionsTests.swift
```diff
@@ -30,7 +30,7 @@ struct MQTTSubscriptionsTests {
 
         var subscriptionID: UInt32 = 0
 
-        switch try subscriptions.addSubscription(continuation: continuation, subscriptions: subscribeInfos, version: .v3_1_1) {
```
adam-fowler, line 33: Can we add some tests in here testing sessions across multiple connections

adam-fowler (commented):
It looking quite good.

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
```diff
@@ -420,13 +435,12 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
...
     }
 
-    private func failTasksAndCloseSubscriptions(with error: any Error) {
+    private func failTasks(with error: any Error) {
         switch self.stateMachine.close() {
         case .failTasksAndClose(let tasks):
             for task in tasks {
                 task.fail(error)
             }
-            self.subscriptions.close(error: error)
```
adam-fowler, line 429: What happens to subscriptions that were made through `MQTTConnection.subscribe`? Should these not return an error when the connection fails.

fpseverino: Yeah, I felt there was something I was missing, thanks

fpseverino: Fixed in https://github.com/swift-server-community/mqtt-nio/pull/222/commits/6e0661d99b7f718f5123a5db0befd27a1996e4e3

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -368,6 +373,29 @@ public final actor MQTTConnection: Sendable {
...
+                let packet = MQTTSubscribePacket(
+                    subscriptions: queuedSubscription.subscriptions,
+                    properties: queuedSubscription.properties,
+                    packetId: self.updatePacketId()
+                )
+                self.channelHandler.subscribe(
+                    id: queuedSubscription.id,
+                    streamContinuation: queuedSubscription.continuation,
+                    packet: packet,
+                    promise: .forget,  // TODO: fix this
```
fpseverino: This could be the source of the timing issues; do you think it’s right to do something like this?

```swift
let promise = self.channel.eventLoop.makePromise(of: UInt32.self)
self.channelHandler.subscribe(..., promise: .nio(promise), ...)
do {
    _ = try await promise.futureResult.get()
} catch {
    queuedSubscription.continuation.finish(throwing: error)
}
```

adam-fowler: Yeah you can't forget the results of the subscribes, given they can fail. I think you'll still get the timing issues in the tests as the handle subscription requests runs concurrently to the withConnection handler

You could do something like above for sure, would be nice it you could queue all the subscribes and wait for their results in a second loop, so all subscribes run in parallel.

adam-fowler (commented):
I think we need some more tests
- Check subscriptions made through MQTTConnection are removed when connection closes. Test both client and server instigating the close
- Check MQTTSession.withSubscription while session is already connected to a connection.
- Check the error from a failed subscription associated with a `MQTTSession.withSubscription` is propagated back to the function

adam-fowler:
Once this is merged I think it would be good to add a new PR that adds a command to MQTTConnection that waits until there are no subscriptions to serve. eg
```swift
MQTTConnection.withConnection(address: .hostname("127.0.0.1"), session: session, logger: logger) { connection, _ in
    try await waitUntilNoActiveSubscriptions()
}
```

Tests/IntegrationTests/IntegrationTests.swift
```diff
@@ -1009,10 +994,78 @@ struct IntegrationTests {
...
+            #expect(!sessionPresent)
+
+            try await withThrowingTaskGroup { group in
+                group.addTask {
+                    try await connection.subscribe(to: [.init(topicFilter: identifier, qos: .atLeastOnce)]) { subscription in
+                        for try await _ in subscription {}
+                    }
+                }
+
+                try await Task.sleep(for: .milliseconds(100))
```
adam-fowler: There is too much reliance on sleep in the tests. This test doesn't need it.
- For the first sleep, once you are inside the closure for a MQTTConnection.subscribe you can guarantee the subscription is setup (unfortunately you can't do this with MQTTSession.subscribe) so you can use an AsyncStream to wait until you are inside the closure.
- The second sleep you just need to wait until the child task exits 
```swift
await #expect(throws: MQTTError.self) {
    try await group.next()
}
```

adam-fowler (commented):
Overall looks good, one small thing with the tests over using sleep. I pointed out a way to get around it in the one test. It might be worthwhile checking the other tests. If you cannot remove the sleep please comment why.

Also can you edit the commit messages where you have used `Copilot`. To include `Assisted-by: Copilot <copilot@github.com>` instead of `Co-authored-by`. `Co-authored-by` implies Copilot has some responsibility for this PR. It doesn't. Only you have responsibility for the committed code. This is something we have asked in Hummingbird and I'd like to propagate it across all of my projects.

fpseverino:
I have updated the commit messages as requested.
I have also reviewed every single instance of `Task.sleep` in the new tests, updated the `connectionSubscriptionCleanup` test as you suggested, removed all unnecessary sleeps, and ensured that the remaining ones are all commented (unfortunately they are all needed to wait for the `session.subscribe` subscription to be set up)

adam-fowler (approved):

PR #223, opened 2026-02-24
Wait for channel to close when exiting `withConnection`

ptoffy:
To follow structured concurrency we should ensure that connection resources are cleaned up before exiting the `withConnection` closure since the channel's lifetime is over

### DIFF
--- a/Sources/MQTTNIO/Connection/MQTTConnection.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnection.swift
@@ -99,11 +99,20 @@ public final actor MQTTConnection: Sendable {
         )
         do {
             let value = try await operation(connection, sessionPresent)
-            await connection.close()
+            await connection.closeAndCleanup()
             return value
         } catch {
-            await connection.close()
+            await connection.closeAndCleanup()
             throw error
+        }
+    }
+
+    private func closeAndCleanup() async {
+        self.close()
+        do {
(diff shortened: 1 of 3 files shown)


Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -367,7 +373,7 @@ public final actor MQTTConnection: Sendable {
     }
 
     /// Close connection
-    public nonisolated func close() {
+    public func close() async {
```
adam-fowler: You cannot remove nonisolated from this function. If you want to wait until the connection is closed in withConnection do it in there.

ptoffy: Why not? We need `assumeIsolated` either way

ptoffy: I mean we can leave it there if you prefer, we save an extra actor hop but we need it to be async either way to wait for the channel to close

adam-fowler: If we were ever to use the ConnectionPool we would need it to be non-async

ptoffy: I see, I didn't know that. I pushed an alternative

adam-fowler (approved):
Looks good

PR #226, opened 2026-04-29
Fix Swift 6.3 sendability issues in connection setup

brientim:
## Summary
- mark `MQTTTaskHandler` as `@unchecked Sendable` for event-loop confined use in `@Sendable` closures
- require `@Sendable` for websocket `afterHandlerAdded` callback
- make `TSTLSConfiguration.tlsDispatchQueue` immutable (`static let`)
- avoid capturing a non-Sendable `[ChannelHandler]` array in websocket setup by building handlers inline

## Backward compatibility
- This remains backward compatible for existing runtime behavior and public API usage: no wire protocol behavior changes, no MQTT semantic changes, and no required call-site changes for typical client usage.
- The changes are narrowly scoped to concurrency annotations/safety (`@Sendable`, `@unchecked Sendable`) and immutable static state, which are needed to satisfy Swift 6 strict-concurrency checks while preserving pre-Swift-6 behavior.

## Validation
- This patch unblocks Swift 6 strict concurrency errors seen from downstream app builds.
- Local `swift test` requires a running MQTT broker on localhost and fails without one (connection refused), so compile verification is based on downstream integration builds.

### DIFF
--- a/Sources/MQTTNIO/ChannelHandlers/MQTTTaskHandler.swift
+++ b/Sources/MQTTNIO/ChannelHandlers/MQTTTaskHandler.swift
@@ -14,7 +14,7 @@
 import NIO
 
 /// Task handler.
-final class MQTTTaskHandler: ChannelInboundHandler, RemovableChannelHandler {
+final class MQTTTaskHandler: @unchecked Sendable, ChannelInboundHandler, RemovableChannelHandler {
     typealias InboundIn = MQTTPacket
 
     var eventLoop: EventLoop!
--- a/Sources/MQTTNIO/MQTTConnection.swift
+++ b/Sources/MQTTNIO/MQTTConnection.swift
@@ -57,11 +57,6 @@ final class MQTTConnection {
                 .channelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: 1)
                 .connectTimeout(client.configuration.connectTimeout)
                 .channelInitializer { channel in
-                    // Work out what handlers to add
-                    let handlers: [ChannelHandler] = [
-                        MQTTMessageHandler(client, pingInterval: pingInterval),
-                        taskHandler,
-                    ]
                     // are we using websockets
                     if let webSocketConfiguration = client.configuration.webSocketConfiguration {
                         // prepare for websockets and on upgrade add handlers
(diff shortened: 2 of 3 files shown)


Sources/MQTTNIO/MQTTConnection.swift
```diff
@@ -75,10 +70,14 @@ final class MQTTConnection {
                             webSocketConfiguration: webSocketConfiguration,
                             upgradePromise: promise
                         ) {
-                            try channel.pipeline.syncOperations.addHandlers(handlers)
+                            try channel.pipeline.syncOperations.addHandler(MQTTMessageHandler(client, pingInterval: pingInterval))
```
adam-fowler, line 73: Any reason you add these individually here, but lower down do them with one call?

Sources/MQTTNIO/MQTTConnection.swift
```diff
@@ -75,10 +70,14 @@ final class MQTTConnection {
...
+                            try channel.pipeline.syncOperations.addHandler(taskHandler)
                         }
                     } else {
-                        return channel.pipeline.addHandlers(handlers)
+                        return channel.pipeline.addHandlers([
```
adam-fowler, line 77: You could also do 
```swift
return channel.eventLoop.makeCompletedFuture {
    try channel.pipeline.syncOperations.addHandlers([
        MQTTMessageHandler(client, pingInterval: pingInterval),
        MQTTTaskHandler(client: client),
    ])
}
```
This wouldn't require MQTTTaskHandler to be Sendable

Sources/MQTTNIO/ChannelHandlers/MQTTTaskHandler.swift
```diff
@@ -14,7 +14,7 @@
 import NIO
 
 /// Task handler.
-final class MQTTTaskHandler: ChannelInboundHandler, RemovableChannelHandler {
+final class MQTTTaskHandler: @unchecked Sendable, ChannelInboundHandler, RemovableChannelHandler {
```
adam-fowler, line 17: I'd rather we don't add more `@unchecked Sendable` to the project.

adam-fowler (commented):
Added some comments. You can do this without making `MQTTTaskHandler` `@unchecked Sendable`.

Don't create the taskHandler in `create`, create it when you add it as a handler. Then change `create` to this.
```swift
self.createBootstrap(client: client, pingInterval: pingInterval)
    .flatMapThrowing { channel in
        let taskHandler = try channel.pipeline.syncOperations.handler(type: MQTTTaskHandler.self)
        return MQTTConnection(channel: channel, cleanSession: cleanSession, timeout: client.configuration.timeout, taskHandler: taskHandler)
    }
```

Just checked the above a little more work is needed as the EventLoop the bootstrap is running on has to be the same EventLoop that the initial promise created in `createBootstrap` is running on. You could pass the `eventLoop` to `getBootstrap` and replace any occurrences of `client.eventLoopGroup` with the passed in EventLoop.

As an aside we are doing a complete rewrite that is completely structured concurrency based.

brientim:
The approach is viable and removes the need for `@unchecked Sendable` while keeping strict-concurrency compilation green. This has been verified and tested, and the PR could be closed.

brientim:
@adam-fowler thank you for the quick reply. The solution provided removes the need for @unchecked Sendable while keeping strict-concurrency compilation green. I am closing this PR.

brientim:
Defect report update: this PR is now closed.

PR #227, opened 2026-04-29
Improve ping configuration API

fpseverino:
- Use `Duration` instead of `TimeAmount` in all public API
- Replace `disablePing` and `pingInterval` with new `PingConfiguration` enum


fpseverino:
CI failures will be fixed once we merge #222

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
```diff
@@ -157,18 +157,26 @@ public struct MQTTConnectionConfiguration: Sendable {
         public var initialRequestHeaders: HTTPHeaders
     }
 
+    /// Configuration for sending `PINGREQ` messages.
+    public enum PingConfiguration: Sendable {
```
adam-fowler: This should be a struct, so we can add additional cases later without breaking changes

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
```diff
@@ -62,7 +61,7 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
 
     private func setInitialized(context: ChannelHandlerContext) {
         self.stateMachine.setInitialized(context: context)
-        if !self.configuration.disablePing {
+        if self.pingreqTimeout != nil {
```
adam-fowler: You check pingreqTimeout being non-nil inside the function as well, we should only need one check

adam-fowler (commented):
In general pretty good, just a couple of minor issues

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
fpseverino: The problem is that we cannot pass the same `pingreqTimeout` from `setInitialized` to `schedulePingreqCallback` and have `schedulePingreqCallback` recursively pass the same value back to itself, because as soon as the CONNACK arrives, `MQTTChannelHandler.pingreqTimeout` might be updated.

So I’ve left the check in both `setInitialized` and `schedulePingreqCallback`, but I’ve removed it from inside the `NIOScheduledCallbackHandler`, as that might actually be redundant.
Please let me know if you think [this](https://github.com/swift-server-community/mqtt-nio/pull/227/commits/95ec1f21382ba4a8ef1b1ddf9037a3fef0394d34) is correct

adam-fowler: Why can't you remove the check from `setInitialized`. It would call `schedulePingreqCallback` and jump out immediately because `self.pingreqTimeout` is nil.

adam-fowler (commented):
One last question otherwise we are good to go

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
fpseverino: Yeah, of course, you're right 🤦‍♂️

adam-fowler (approved):

PR #230, opened 2026-05-08
Implement `waitUntilNoActiveSubscriptions`

fpseverino:
Adds a method on `MQTTSession` that waits until there are no active subscriptions opened via the session or via any `MQTTConnection`s using the session.

I implemented it on `MQTTSession` since IMHO that is its natural place from a code perspective, but we can easily add a wrapper on `MQTTConnection` if you prefer.

### DIFF
--- a/Sources/MQTTNIO/Session/MQTTSession.swift
+++ b/Sources/MQTTNIO/Session/MQTTSession.swift
@@ -138,8 +138,8 @@ extension MQTTSession {
                 }
                 self.logger.trace("Waiting for \(subscriptions.subscriptionIDMap.count) active subscriptions to complete")
                 subscriptions.emptySubscriptionsContinuations.append(continuation)
-                self.logger.trace("All active subscriptions completed")
             }
         }
+        self.logger.trace("All active subscriptions completed")
     }
 }
--- a/Sources/MQTTNIO/Session/MQTTSession.swift
+++ b/Sources/MQTTNIO/Session/MQTTSession.swift
@@ -134,6 +134,7 @@ extension MQTTSession {
             self.subscriptions.withLock { subscriptions in
                 if subscriptions.subscriptionIDMap.isEmpty {
                     self.logger.trace("No active subscriptions to wait for")
+                    continuation.resume()
                     return
                 }
                 self.logger.trace("Waiting for \(subscriptions.subscriptionIDMap.count) active subscriptions to complete")
--- a/Sources/MQTTNIO/Session/MQTTSession.swift
+++ b/Sources/MQTTNIO/Session/MQTTSession.swift
@@ -28,9 +28,14 @@ public final class MQTTSession: Sendable {
 
     let subscriptions: Mutex<MQTTSubscriptions>
 
+    /// See ``MQTTSession/waitUntilNoActiveSubscriptions()``
+    let emptySubscriptionsStream: AsyncStream<Void>
+
     let subscriptionsQueue: AsyncStream<SessionSubscriptionTask>
     @usableFromInline
     let subscriptionsQueueContinuation: AsyncStream<SessionSubscriptionTask>.Continuation
+
+    let logger: Logger
 
     /// Initialize a new ``MQTTSession`` with a unique client identifier.
     ///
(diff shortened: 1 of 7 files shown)


Sources/MQTTNIO/Session/MQTTSession.swift
```diff
@@ -118,4 +126,21 @@ extension MQTTSession {
...
+    /// so for example if a subscription is opened via this ``MQTTSession`` but the session is not passed to a ``MQTTConnection``,
+    /// this function will not wait for that subscription.
+    ///
+    /// If new subscriptions are opened while waiting, this function will also wait for those subscriptions to complete.
+    public func waitUntilNoActiveSubscriptions() async {
+        if self.subscriptions.withLock({ $0.subscriptionIDMap.isEmpty }) {
+            self.logger.trace("No active subscriptions to wait for")
+            return
+        }
+        self.logger.trace("Waiting for \(self.subscriptions.withLock { $0.subscriptionIDMap.count }) active subscriptions to complete")
```
fpseverino: These log messages were useful when writing tests, if you prefer we can remove them

adam-fowler (commented):
Couple of things
- While this is really about the number of subscriptions on the session I think this should be a public function on MQTTConnection. Although internally it will probably be running wherever the subscriptions are stored.
- Secondly you can't really do this with an AsyncStream. If you have two functions waiting only one would return when the sub count goes to zero. You'll need to do it as an array of continuations (protected by a lock) which are all completed when the sub count reaches zero. You'll also have to be careful about when you add new continuations that you don't add them after the subscriptions have emptied. You can't check you have subscriptions and then add a continuation as the subscription list could empty in between those two operations. So you have to make that an atomic operation.

fpseverino:
> So you have to make that an atomic operation.

All of the critical `waitUntilNoActiveSubscriptions` code is now inside `self.subscriptions.withLock`, is that sufficient?

> I think this should be a public function on MQTTConnection.

I added a wrapper method in `MQTTConnection`. Should we keep the original `MQTTSession` `public` or make it `internal`?

Sources/MQTTNIO/Session/MQTTSession.swift
```diff
@@ -118,4 +121,25 @@ extension MQTTSession {
...
+    /// so for example if a subscription is opened via this ``MQTTSession`` but the session is not passed to a ``MQTTConnection``,
+    /// this function will not wait for that subscription.
+    ///
+    /// If new subscriptions are opened while waiting, this function will also wait for those subscriptions to complete.
+    public func waitUntilNoActiveSubscriptions() async {
+        await withCheckedContinuation { continuation in
+            self.subscriptions.withLock { subscriptions in
+                if subscriptions.subscriptionIDMap.isEmpty {
+                    self.logger.trace("No active subscriptions to wait for")
+                    return
```
adam-fowler, line 138: You need to call `continuation.yield()` here, before returning

Tests/MQTTNIOTests/MQTTConnectionTests.swift
```diff
@@ -802,24 +786,58 @@ struct MQTTConnectionTests {
...
-            // send SUBACK
-            let unsuback = MQTTSubAckPacket(
-                type: .UNSUBACK,
-                packetId: unsubscribePacket.packetId,
-                reasons: subscribeInfos.map { _ in .success },
-                properties: .init()
-            )
-            try await channel.writeInboundPacket(unsuback, version: .v3_1_1)
+    @Test("Wait Until No Active Subscriptions")
+    func waitUntilNoActiveSubscriptions() async throws {
```
adam-fowler, line 798: Can we add a test for when `connection.waitUntilNoActiveSubscriptions` returns when there were no subscriptions already.

fpseverino: I’ve added a very simple test, but it wouldn’t have worked without resuming the continuation before returning, as you suggested

Sources/MQTTNIO/Session/MQTTSession.swift
adam-fowler: or resume 😄

Tests/MQTTNIOTests/MQTTConnectionTests.swift
adam-fowler: That's why I suggested adding the test

adam-fowler (approved):

PR #233, opened 2026-05-25
Session storage

### LINKED ISSUE #225: v3: Create  non-copyable type SessionStorage that is passed between session and connection
Create non-copyable type SessionStorage which contains inflightPackets and subscriptions (and perhaps the ID). MQTTSession has an optional reference to this. When a connection is made it is passed to the MQTTConnection and when the connection closes it passed back to MQTTSession. With this we should be able to remove the Mutex access to packets and subscriptions

adam-fowler:
Separate subscription state, inflight packets and client ID from session and store in separate type `MQTTSessionStorage`. Which is stored inside a `UniqueReference` in the MQTTSession. When a connection is created that is serving the session it borrows the session storage and stores it in the channel handler until the connection closes. On close it will update the session storage in the MQTTSession with the state from the channel handler.

If another connection attempts to borrow the session state while it is already borrowed it will throw an error saying this session is already being served by a connection.

This change does mean it is not possible to access various session parameters from `MQTTSession`.

### DIFF
--- a/Sources/MQTTNIO/Connection/MQTTConnection.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnection.swift
@@ -45,15 +45,9 @@ public final actor MQTTConnection: Sendable {
     let channelHandler: MQTTChannelHandler
     let configuration: MQTTConnectionConfiguration
     let globalPacketId = Atomic<UInt16>(1)
-    /// Local copy of inflight messages
-    /*private var inflight: MQTTInflight {
-        get { self.channelHandler.sessionStorage.inflightPackets }
-        set { self.channelHandler.sessionStorage.inflightPackets = newValue }
-    }*/
     let isClosed: Atomic<Bool>
 
     var connectionParameters = ConnectionParameters()
-    //let session: MQTTSession
 
     /// Initialize connection
     private init(
--- a/Sources/MQTTNIO/Connection/MQTTConnection.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnection.swift
@@ -100,22 +100,23 @@ public final actor MQTTConnection: Sendable {
         )
         do {
             let value = try await operation(connection, sessionPresent)
-            await connection.closeAndCleanup(session: nil)
+            await connection.closeAndCleanup()
             return value
         } catch {
-            await connection.closeAndCleanup(session: nil)
+            await connection.closeAndCleanup()
             throw error
         }
     }
 
-    package func closeAndCleanup(session: MQTTSession?) async {
+    @discardableResult package func closeAndCleanup() async -> MQTTSessionStorage {
         self.close()
-        session?.storage.put(self.channelHandler.session)
+        let sessionStorage = self.channelHandler.session
         do {
             try await self.channel.closeFuture.get()
         } catch {
             logger.error("Failed to close connection", metadata: ["error": "\(error)"])
         }
+        return sessionStorage
     }
 
     /// Connect to MQTT server with a clean session and run operations using the connection and then close it.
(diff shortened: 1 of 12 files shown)


Sources/MQTTNIO/Session/MQTTSession.swift
```diff
@@ -18,15 +18,8 @@ import Synchronization
...
-    private let _clientID: Mutex<String>
-
-    /// Whether a ``MQTTConnection`` is currently connected using this session.
-    let isConnected: Atomic<Bool>
-
-    /// Inflight messages
-    let inflightPackets: Mutex<MQTTInflight>
-
-    let subscriptions: Mutex<MQTTSubscriptions>
+    // data associated with session that is borrowed by the connection
```
fpseverino:
```suggestion
    /// Data associated with session that is borrowed by the connection
```

Sources/MQTTNIO/Utils/UniqueReference.swift
```diff
@@ -0,0 +1,63 @@
...
+
+import Synchronization
+
+enum UniqueReferenceError: Error {
+    case alreadyBorrowed
+}
+
+struct UniqueReference<Value: Sendable>: ~Copyable, Sendable {
+
+    let ref: Mutex<Value?>
```
fpseverino, line 21:
```suggestion
struct UniqueReference<Value: Sendable>: ~Copyable, Sendable {
    let ref: Mutex<Value?>
```

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -364,22 +383,16 @@ public final actor MQTTConnection: Sendable {
             properties: properties,
             will: publish
         )
-        guard
+        /*guard
             self.session.isConnected.compareExchange(expected: false, desired: true, successOrdering: .relaxed, failureOrdering: .relaxed).exchanged
         else {
             throw MQTTError.alreadyConnectedWithSession
-        }
+        }*/
```
fpseverino: We don’t need this anymore, do we?

Sources/MQTTNIO/Utils/UniqueReference.swift
```diff
@@ -0,0 +1,63 @@
...
+
+    init(_ value: Value) {
+        self.ref = .init(value)
+    }
+
+    /// borrow value and return a mutated version
+    func mutatingBorrow<Return>(_ operation: (Value) async -> (Value, Result<Return, any Error>)) async throws -> Return {
+        guard let value = take() else { throw UniqueReferenceError.alreadyBorrowed }
+        let (newValue, result) = await operation(value)
```
fpseverino, line 30: If the Task is cancelled during `operation`, that’s not a problem, right?
`put` is called anyway, isn't it?

fpseverino (approved):
LGTM, just a few formatting notes and a doubt about Task cancellation within `mutatingBorrow`

fpseverino:
I think `UniqueReference` is an excellent idea, and with a different name for the type that doesn’t bring to mind reference types, and different method names that don’t evoke `~Copyable` types (such as `borrow` and the like), it could be a valuable addition to the Swift standard library or a package like AsyncAlgorithms

Sources/MQTTNIO/Connection/MQTTConnection.swift
adam-fowler: no we don't

Sources/MQTTNIO/Utils/UniqueReference.swift
adam-fowler: A cancellation can only be returned as a `Result.failure` so the value would still be put back

PR #232, opened 2026-05-23
Update docs for v3

fpseverino:
Resolves #229

PR #239, opened 2026-05-28
Add waitOnClose, clean up withConnection calls

adam-fowler:
The connection being closed by the server is currently only known when the client performs some action or the connection is waiting on a subscription. If the client is doing other work this will not be cancelled. Adding a waitOnClose will allow users to catch this situation.

I also decided to remove two of the withConnection functions to clean things up a little.


fpseverino (approved):
LGTM, I also removed one of the `withConnection` functions in #232, let's see how GitHub handles that

PR #240, opened 2026-05-28
Clean start on empty session, Don't cancel MQTTSession.subscriptionsQueue iteration

### LINKED ISSUE #236: v3: Set session clean start on empty session
If the MQTT client is v5 and the session passed in is empty ie no stored subscriptions, or inflight packets then set clean start to true.

Unfortunately we can't do the same for v3.1.1 as this would mean the session wouldn't be saved on disconnection

adam-fowler:
This PR was meant to be simple. It would start a session with cleanStart if we had a v5 connection and the session provided was empty. This meant I had to change the SessionExpiryInterval test to add a subscription to the session otherwise it'd keep starting with a clean session. But....

I couldn't get the session subscription to work and this was because when I ended the first connection it cancelled the MQTTSession.subscription iteration, which ends the stream. So any subsequent session subscription requests were ignored. So I have to ensure iterating MQTTSession.subscription is never cancelled.

First I had to have a way to end iterating it, so I added a new request `.cancelled`, which would stop iterating the queue. Secondly I had to protect the iteration from receiving a task cancellation. Currently the only method to do this is run your code in an unstructured task so that is what I've done. When Swift 6.4 comes out I can use `withTaskCancellationShield`

### DIFF
--- a/Sources/MQTTNIO/Connection/MQTTConnection.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnection.swift
@@ -281,7 +281,23 @@ public final actor MQTTConnection: Sendable {
             }
         let connection = try await future.get()
         try await connection.waitOnInitialized()
-        let sessionPresent = try await connection.sendConnect(clientID: _session.clientID, cleanSession: session == nil)
+
+        // cleanSession means different things for v3.1.1 and v5.0. If you set cleanSession in v3.1.1 it will
+        // also not save the session, but you can set cleanSession (called cleanStart in v5) in v5 as it
+        // will save the session if the session expiry interval is set. Therefore we can start with clean start
+        // if our session is empty on v5
+        let cleanSession =
+            if let session {
+                switch configuration.version {
+                case .v3_1_1:
+                    false
+                case .v5_0:
+                    session.subscriptions.subscriptionIDMap.isEmpty && session.inflight.packets.isEmpty
+                }
+            } else {
+                true
+            }
+        let sessionPresent = try await connection.sendConnect(clientID: _session.clientID, cleanSession: cleanSession)
         return (connection, sessionPresent)
     }
 
@@ -302,7 +318,7 @@ public final actor MQTTConnection: Sendable {
             case .v3_1_1(let will):
                 will.map {
                     MQTTPublishInfo(
-                        qos: .atMostOnce,
+                        qos: $0.qos,
                         retain: $0.retain,
                         dup: false,
                         topicName: $0.topicName,
@@ -313,7 +329,7 @@ public final actor MQTTConnection: Sendable {
             case .v5_0(_, _, let will, _):
                 will.map {
                     MQTTPublishInfo(
-                        qos: .atMostOnce,
+                        qos: $0.qos,
                         retain: $0.retain,
                         dup: false,
                         topicName: $0.topicName,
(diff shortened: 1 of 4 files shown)


Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -281,7 +285,23 @@ public final actor MQTTConnection: Sendable {
...
+                switch configuration.version {
+                case .v3_1_1:
+                    false
+                case .v5_0:
+                    session.subscriptions.subscriptionIDMap.isEmpty && session.inflight.packets.isEmpty
+                }
+            } else {
+                true
+            }
+        let sessionPresent = try await connection.sendConnect(clientID: _session.clientID, cleanSession: cleanSession)
```
adam-fowler, line 304: This change was meant to the whole of the PR.

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -168,15 +168,19 @@ public final actor MQTTConnection: Sendable {
                     return (sessionStorage, .failure(error))
                 }
                 if !sessionPresent { await connection.closeSubscriptions() }
+
+                // stick this in an unstructured task to avoid cancellation on iterating the subscribe
+                // request queue. Will use `withTaskCancellationShield` when Swift 6.4 comes out
```
```suggestion
                // TODO: stick this in an unstructured task to avoid cancellation on iterating the subscribe
                // request queue. Will use `withTaskCancellationShield` when Swift 6.4 comes out
```
Let's add a TODO so we don't forget to change it when Swift 6.4 comes out

Tests/IntegrationTests/IntegrationV5Tests.swift
```diff
@@ -263,53 +262,66 @@ struct IntegrationV5Tests {
...
-            // but the Session Expiry Interval was set to 0
-            #expect(sessionPresent == false)
-            try await connection.ping()
-        }
+        try await withThrowingTaskGroup { group in
+            group.addTask {
+                try await session.subscribe(to: [.init(topicFilter: "test", qos: .atMostOnce)]) { sub in
+                    var iterator = sub.makeAsyncIterator()
+                    _ = try? await iterator.next()
+                }
```
fpseverino, line 271: This is added to avoid sending `cleanStart` in subsequent connections, right?

Tests/MQTTNIOTests/Utils/MQTTConnectionTests+withTest.swift
```diff
@@ -62,8 +62,8 @@ extension MQTTConnectionTests {
                         do {
                             _ = try await connection.sendConnect(clientID: sessionStorage.clientID, cleanSession: sessionStorage.clientID.isEmpty)
                             try await withThrowingTaskGroup { group in
-                                group.addTask { try await connection.handleSessionSubscriptionTasks(session: session) }
-                                defer { group.cancelAll() }
+                                group.addTask { await connection.handleSessionSubscriptionTasks(session: session) }
+                                defer { session.subscriptionsQueueContinuation.yield(.cancel) }
```
fpseverino: Why don’t we need the unstructured Task here?

fpseverino (approved):
LGTM, just a couple of curiosities

Sources/MQTTNIO/Connection/MQTTConnection.swift
adam-fowler: Ok But I'd move the TODO: next to the actual action

Tests/IntegrationTests/IntegrationV5Tests.swift
adam-fowler: Yes that is right, I'll add a comment

Tests/MQTTNIOTests/Utils/MQTTConnectionTests+withTest.swift
adam-fowler: I'll fix this one as well

adam-fowler: done

PR #243, opened 2026-05-29
Fix edge case in WebSocket initial request handler

adam-fowler:
Fix edge case in WebSocketInitialRequestHandler where the handler becomes active before it is added


Sources/MQTTNIO/ChannelHandlers/WebSocketInitialRequest.swift
```diff
@@ -33,7 +33,18 @@ final class WebSocketInitialRequestHandler: ChannelInboundHandler, RemovableChan
         self.upgradePromise = upgradePromise
     }
 
+    public func handlerAdded(context: ChannelHandlerContext) {
```
fpseverino: There are a few `public` declarations in this file that we shouldn't need

fpseverino (approved):
While you're at it, it would be great if you could remove all the 'public' declarations from this file

PR #246, opened 2026-05-29
Exit waitUntilNoActiveSubscriptions on connection close and cancellation

### LINKED ISSUE #238: v3: MQTTConnection.waitUntilNoActiveSubscriptions should error if connection is closed

adam-fowler:
- waitUntilNoActiveSubscriptions now throws
- If connection is already closed, pass connectionClosed error to continuation 
- When MQTTSubscriptions.close(error:) is called pass the error to the stored continuations.
- Add an id with each continuation so it can be resumed by id. Needed to implement cancellation
- Add a test to check function returns in all situations

### DIFF
--- a/Sources/MQTTNIO/Connection/MQTTConnection.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnection.swift
@@ -225,17 +225,37 @@ public final actor MQTTConnection: Sendable {
     /// This function waits only for subscriptions that have been acknowledged by the server.
     ///
     /// If new subscriptions are opened while waiting, this function will also wait for those subscriptions to complete.
-    public func waitUntilNoActiveSubscriptions() async {
-        await withCheckedContinuation { continuation in
-            if self.channelHandler.session.subscriptions.subscriptionIDMap.isEmpty {
-                self.logger.trace("No active subscriptions to wait for")
-                continuation.resume()
-                return
+    ///
+    /// - Throws: If connection is closed while waiting it will throw the error that caused the connection to close
+    public func waitUntilNoActiveSubscriptions() async throws {
+        let id = Self.requestIDGenerator.next()
+        try await withTaskCancellationHandler {
+            try await withCheckedThrowingContinuation { continuation in
+                if self.channelHandler.session.subscriptions.subscriptionIDMap.isEmpty {
+                    self.logger.trace("No active subscriptions to wait for")
+                    continuation.resume()
+                    return
+                }
+                if self.channelHandler.stateMachine.isClosed() {
+                    continuation.resume(throwing: MQTTError.connectionClosed)
+                    return
+                }
+                self.logger.trace("Waiting for \(self.channelHandler.session.subscriptions.subscriptionIDMap.count) active subscriptions to complete")
+                self.channelHandler.session.subscriptions.addWaitForEmptySubscriptions(id: id, continuation: continuation)
             }
-            self.logger.trace("Waiting for \(self.channelHandler.session.subscriptions.subscriptionIDMap.count) active subscriptions to complete")
-            self.channelHandler.session.subscriptions.emptySubscriptionsContinuations.append(continuation)
+        } onCancel: {
+            self.cancelWaitForEmptySubscriptions(id: id)
         }
         self.logger.trace("All active subscriptions completed")
+    }
+
+    @usableFromInline
+    nonisolated func cancelWaitForEmptySubscriptions(id: Int) {
+        self.channel.eventLoop.execute {
+            self.assumeIsolated { this in
+                this.channelHandler.session.subscriptions.cancelWaitForEmptySubscriptions(id: id)
+            }
+        }
     }
 
     /// Connect to MQTT server and return the connection and whether there was a previous session present.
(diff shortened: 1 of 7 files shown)


Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -225,19 +225,39 @@ public final actor MQTTConnection: Sendable {
...
+                }
+                self.logger.trace("Waiting for \(self.channelHandler.session.subscriptions.subscriptionIDMap.count) active subscriptions to complete")
+                self.channelHandler.session.subscriptions.addWaitForEmptySubscriptions(id: id, continuation: continuation)
             }
-            self.logger.trace("Waiting for \(self.channelHandler.session.subscriptions.subscriptionIDMap.count) active subscriptions to complete")
-            self.channelHandler.session.subscriptions.emptySubscriptionsContinuations.append(continuation)
+        } onCancel: {
+            self.cancelWaitForEmptySubscriptions(id: id)
         }
         self.logger.trace("All active subscriptions completed")
```
fpseverino, line 249: This is a problem that existed previously, but this log message is written even when there are no subscriptions to wait for; it would be better not to write it in that case

Tests/MQTTNIOTests/Utils/MQTTConnectionTests+withTest.swift
fpseverino, line 1: These changes "break" the `MQTTConnectionTests.waitUntilNoActiveSubscriptions` test, because whilst we are waiting for the `Task.sleep`, the PUBLISH packet has already been sent and the subscriptions have ended, so `connection.waitUntilNoActiveSubscriptions()` has nothing to wait for and returns immediately. If we want to simplify the helper, we can move the AsyncStream used in the test directly into the helper, something like this:
```swift
    func withTestSessionSubscriptions(...) async throws {
        let (stream, continuation) = AsyncStream.makeStream(of: Void.self)
        try await withTestMQTTServer(session: session, logger: logger) { connection in
            await withThrowingTaskGroup { group in
                group.addTask {
                    // Wait until all subscriptions are active
                    await stream.first { _ in true }

                    try await clientOperation(connection)
                }
                for subscribeInfo in subscribeInfos { /* setup subscription */ }
            }
        } server: { channel in
            for _ in subscribeInfos { /* Receive SUBSCRIBE and send SUBACK */ }

            // Signal that all subscriptions are active
            continuation.yield()

            try await serverOperation(channel)

            for _ in subscribeInfos { /* Receive UNSUBSCRIBE and send UNSUBACK */ }
        }
    }
```
That way we can also remove the AsyncStream from the test

Tests/IntegrationTests/IntegrationTests.swift
```diff
@@ -1113,6 +1113,76 @@ struct IntegrationTests {
...
+                    group.addTask {
+                        try await Task.sleep(for: .milliseconds(50))
+                        connection.close()
+                    }
+                    try await connection.waitUntilNoActiveSubscriptions()
+                }
+            }
+        }
+
+        // expect waitUntilNoActiveSubscriptions to throw error as it has been cancelled
```
fpseverino, line 1173: No error is thrown here. We’re testing whether `waitUntilNoActiveSubscriptions` throws an error if the Task is cancelled, right? How can we see that?

Tests/MQTTNIOTests/Utils/MQTTConnectionTests+withTest.swift
adam-fowler: `IntegrationTests.waitUntilNoActiveSubscriptions` wasn't working because I forgot to rename the identifier for the cancelActiveSubscriptionsWait test. Are you sure that isn't what you were seeing.

I changed the withTestSession.. function because you need all the subscriptions to have run before the clientOperation is run, otherwise the calls from both could get mixed up.

Tests/IntegrationTests/IntegrationTests.swift
adam-fowler: Fixed, main thing was it exited

Tests/MQTTNIOTests/Utils/MQTTConnectionTests+withTest.swift
adam-fowler: Ok, I've changed it. I don't cancel the subscription tasks as soon as the client operation returns. I've also added a stream to make sure the subscriptions have been setup before the client operation is called

Tests/IntegrationTests/IntegrationTests.swift
```diff
@@ -1178,9 +1178,19 @@ struct IntegrationTests {
         ) { connection, sessionPresent in
             try await connection.publish(to: "test", payload: .init(), qos: .atLeastOnce)
             await stream.first { _ in true }
-            async let _ = connection.waitUntilNoActiveSubscriptions()
+            let (stream2, cont2) = AsyncStream.makeStream(of: Void.self)
```
fpseverino, line 1181: I don't think we need this stream, the `Task.sleep` is necessary and sufficient

Tests/MQTTNIOTests/MQTTConnectionTests.swift
fpseverino: We can remove this stream now

fpseverino (approved):
LGTM, just a couple of nits

PR #249, opened 2026-05-30
Close connection if send of CONNECT packet fails

### LINKED ISSUE #248: v3: If send CONNECT packet fails we should close the connection.
If the sending of the CONNECT packet failed we never close the connection.

adam-fowler:


Tests/IntegrationTests/IntegrationTests.swift
```diff
@@ -43,6 +43,24 @@ struct IntegrationTests {
         }
     }
 
+    @Test("Close On Failed Connect Send")
+    func closeOnFailedConnectSend() async throws {
+        await withThrowingTaskGroup { group in
```
adam-fowler, line 71: It's not easy to check that the connection was closed, but we can at least check it doesn't crash

adam-fowler: So I'm checking logger output to see if connection is closed

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -431,14 +437,16 @@ public final actor MQTTConnection: Sendable {
     }
 
     /// Close connection
-    public nonisolated func close() {
+    public nonisolated func close(sendDisconnect: Bool = true) {
```
fpseverino: Genuine question: do we want to give the user the option to not send DISCONNECT?
If so I think we should document it in the DocC comment

adam-fowler: Oh shit this is a public function, no we don't.

adam-fowler: fixed

fpseverino (approved):

PR #260, opened 2026-06-05
Add Swift Configuration support

### LINKED ISSUE #256: Add swift-configuration support
Add `MQTTConnection.withConnection(configReader: ConfigReader, ...)`
Add `MQTTConnectionConfiguration(configReader: ConfigReader)`

fpseverino:
Resolves #256

- Add an initializer for `MQTTConnectionConfiguration` that uses a `ConfigReader`
- Allow sending of a Password without a User Name (which is allowed in MQTT v5.0, but not in v3.1.1)

### DIFF
--- a/Sources/MQTTNIO/Connection/MQTTConnectionConfiguration+ConfigReader.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnectionConfiguration+ConfigReader.swift
@@ -210,8 +210,10 @@ extension MQTTConnectionConfiguration.TLS {
         )
         tlsConfiguration.trustRoots = nioSSLTrustRoots.map { .certificates($0) }
         self.base = .enable(.niossl(tlsConfiguration), tlsServerName: tlsServerName)
-        #endif
+        return
+        #else
         throw _TLSConfigError.missingConfiguration
+        #endif
     }
 }
 
--- a/Sources/MQTTNIO/Connection/MQTTConnectionConfiguration+ConfigReader.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnectionConfiguration+ConfigReader.swift
@@ -10,11 +10,6 @@ public import Configuration
 import HTTPTypes
 import NIOCore
 
-#if canImport(FoundationEssentials)
-import FoundationEssentials
-#else
-import Foundation
-#endif
 #if canImport(Network)
 import NIOTransportServices
 #endif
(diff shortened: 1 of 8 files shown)


Sources/MQTTNIO/Connection/MQTTConnectionConfiguration+ConfigReader.swift
```diff
@@ -0,0 +1,252 @@
...
+            let tlsConfiguration = TSTLSConfiguration(trustRoots: trustRootsConfig, clientIdentity: privateKeyConfig)
+            self.base = .enable(.ts(tlsConfiguration), tlsServerName: tlsServerName)
+        #endif
+        default:
+            throw TLSConfigError.incompatibleConfiguration
+        }
+    }
+}
+
+struct ConfigHTTPField: ExpressibleByConfigString {
```
fpseverino, line 220: I did this to avoid doing `extension HTTPField: @retroactive ExpressibleByConfigString`.

I'm sure you remember my last adventure with `@retroactive` conformances 😂

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration+ConfigReader.swift
```diff
@@ -0,0 +1,252 @@
...
+    ///
+    /// ## Configuration keys
+    /// - `will.topicName` (string): Topic Name of the Will Message.
+    /// - `will.payload` (bytes): Payload of the Will Message.
+    /// - `will.qos` (int): QoS level of the Will Message.
+    /// - `will.retain` (bool, optional, default: `false`): Specifies if the Will Message is to be retained when it is published.
+    ///
+    /// - Parameter config: The config reader to read configuration values from.
+    init(config: ConfigReader) throws {
+        let willConfig = config.scoped(to: "will")
```
adam-fowler: Can we move the scoping out to `MQTTConnectionConfiguration.init(config:)` and do the same for all the other functions as well.

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration+ConfigReader.swift
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /// Creates a new TLS configuration using values from the provided reader.
+    ///
+    /// ## Configuration keys
+    /// - `tls.config` (string, optional): Whether to use `NIOSSL` or `NIOTransportServices`.
+    /// - `tls.certificateChain` (string): TLS certificate chain in PEM format. Only applicable for `NIOSSL`.
+    /// - `tls.privateKey` (string): TLS private key, in PEM format for NIOSSL, or as a file path to a `.p12` file for NIOTransportServices.
+    /// - `tls.privateKeyPassword` (string): Password for the TLS private key. Only applicable for NIOTransportServices.
+    /// - `tls.trustRoots` (string, optional): TLS trust roots, in PEM format for NIOSSL, or as a file path to a `.der` file for NIOTransportServices.
```
adam-fowler: I'm not a great fan of sharing configuration keys when they point to different things dependent on the setting of another key. Perhaps the TS ones should be scoped to `tls.ts`

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration+ConfigReader.swift
```diff
@@ -0,0 +1,252 @@
...
+    /// - `webSocket.initialRequestHeaders` (string array, optional): Initial HTTP headers to include in the WebSocket handshake request.
+    ///
+    /// - Parameter config: The config reader to read configuration values from.
+    init?(config: ConfigReader) {
+        let webSocketConfig = config.scoped(to: "webSocket")
+        let urlPath = webSocketConfig.string(forKey: "urlPath")
+        let maxFrameSize = webSocketConfig.int(forKey: "maxFrameSize")
+        let initialRequestHeaders: HTTPFields?
+        if let initialRequestHeadersArray = webSocketConfig.stringArray(forKey: "initialRequestHeaders", as: ConfigHTTPField.self) {
+            var headers = HTTPFields()
```
adam-fowler: You can create an HTTPFields from an array of HTTPField, instead of doing the work yourself
```suggestion
            var headers = HTTPFields(initialRequestHeadersArray)
```

fpseverino: The only other available initializer is [this one](https://swiftpackageindex.com/apple/swift-http-types/1.6.0/documentation/httptypes/httpfields/init(parsedtrailerfields:)), which takes already parsed fields, and I have had trouble with it in the past. But in addition, we would in any case need to convert our custom `ConfigHTTPField`s to `HTTPField`s, so some sort of iteration is needed anyway

adam-fowler: You could have ConfigHTTPField wrap an HTTPField
```swift
struct ConfigHTTPField: ExpressibleByConfigString {
    let field: HTTPField
    ...
```
And then use a lazy map
```swift
initialRequestHeaders = HTTPFields(initialRequestHeadersArray.lazy.map { $0.field })
```

fpseverino: TIL about `lazy`!

adam-fowler (approved):

Issue #172, opened 2025-06-12
Any plans to support MQTT over QUIC?


kuekiko:
Hi, thanks for your great work on this library.
I'm wondering if there are any plans to support MQTT over QUIC in future releases?
Given the growing adoption of QUIC and its potential benefits for mobile and unstable networks, this would be a very valuable feature.
Looking forward to your thoughts!

adam-fowler:
Closing this as it is dependent on QUIC being support at NIO level which doesn't look like happening in near future

fpseverino:
[SwiftNIO QUIC](https://github.com/apple/swift-nio-quic) is now available, which will enable us to support MQTT over QUIC.
However, at the moment it is still in active development and does not offer a stable API yet, so we’ll have to wait before releasing support for QUIC.

PR #264, opened 2026-06-22
Future-proof public API for QUIC transport

fpseverino:
> [!IMPORTANT]
> This PR depends on #260 

This PR future-proofs the public API of `MQTTConnectionConfiguration`, so that when we add support for MQTT over QUIC, we can do so without breaking the API.
This is the first step towards #172.

When initializing `MQTTConnectionConfiguration` via a `ConfigReader`, the behaviour remains unchanged.

### DIFF
--- a/Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
@@ -292,12 +292,12 @@ public struct MQTTConnectionConfiguration: Sendable {
     }
 
     /// Whether is using WebSockets for connection.
-    public var useWebSockets: Bool {
+    var useWebSockets: Bool {
         if case .webSocket = self.transport.base { true } else { false }
     }
 
     /// URL Path for WebSocket. Defaults to "/mqtt".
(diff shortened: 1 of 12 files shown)


Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
```diff
@@ -258,23 +279,39 @@ public struct MQTTConnectionConfiguration: Sendable {
...
+        switch self.transport.base {
+        case .tcp(let tls):
+            tls.base
+        case .webSocket(_, let tls):
+            tls.base
+        }
     }
 
     /// Whether is using WebSockets for connection.
     public var useWebSockets: Bool {
```
adam-fowler: Do the websocket configuration values need to be public

adam-fowler (approved):

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
fpseverino: No I don't think so, I'll make them internal

PR #266, opened 2026-06-26
MQTT over QUIC

fpseverino:
Implements [MQTT over QUIC - Single Stream Mode](https://github.com/oasis-tcs/mqtt/blob/main/contributions/EMQX/mqtt-over-quic-cn/mqtt_over_quic_single_stream_CN_1.md) using [SwiftNIO QUIC](https://github.com/apple/swift-nio-quic).

See #172

This is a WIP. Here's a list of things that need addressing:
- SwiftNIO QUIC requires Swift 6.3+, but perhaps it's to early for MQTTNIO
- SwiftNIO QUIC is currently pre-v1.0 and depends on beta versions of Swift Crypto (see `SWIFT_CERTIFICATES_ALLOW_SWIFT_CRYPTO_BETA=1`)
- SwiftNIO QUIC emits _**a lot**_ of logs, we probably need to address that by raising our logs' level or by letting the user specify a different logger for the QUIC layer
- We need to handle the `Channel` of the main UDP connection, not only the QUIC Stream one (does [this](https://github.com/swift-server-community/mqtt-nio/pull/266/commits/c209eeecfe0c7361964d0dd3e5c62052806bdf9a) fix it?)
- EMQX needs to be added to macOS CI
- The `generate-emqx-certs.sh` script was made with AI, it needs to be properly reviewed
- The code needs more refining and documentation
- We can probably look at Simple Multistream Mode, even if it's not being standardized


Sources/MQTTNIO/Connection/MQTTConnection+QUIC.swift
```diff
@@ -0,0 +1,87 @@
...
+extension MQTTConnection {
+    @available(iOS 26, macOS 26, tvOS 26, watchOS 26, visionOS 26, *)
+    static func _makeQUICConnection(
+        address: MQTTServerAddress,
+        configuration: MQTTConnectionConfiguration,
+        session: MQTTSessionStorage,
+        eventLoop: any EventLoop,
+        logger: Logger
+    ) async throws -> MQTTConnection {
+        //eventLoop.assertInEventLoop()
```
adam-fowler: Why is this commented out

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -276,27 +276,43 @@ public final actor MQTTConnection: Sendable {
...
+                    }
                 }
+            connection = try await future.get()
+        case .quic:
+            guard #available(iOS 26, macOS 26, tvOS 26, watchOS 26, visionOS 26, *) else {
+                // TODO: handle gracefully
+                preconditionFailure("QUIC is available only on Apple OS 26+")
             }
-        let connection = try await future.get()
+            connection = try await _makeQUICConnection(
```
adam-fowler, line 309: You'll need to do the `eventLoop.inEventLoop` check as is done with `_makeConnection`

fpseverino: The issue is that, unlike `_makeConnection `, `_makeQUICConnection` is `async` (it uses `swift-nio-quic` APIs that are `async`-only), so I can't run it inside `eventLoop.flatSubmit`.

Should I do the [`promise.completeWithTask`](https://swiftpackageindex.com/apple/swift-nio/main/documentation/niocore/eventlooppromise/completewithtask(_:)) dance or is there a better way?

adam-fowler: Given all the handlers called inside _makeQUICConnection are async I think you can ignore my comments. Would be good to delete the commented out `assertInEventLoop` though

Sources/MQTTNIO/Connection/MQTTConnection+QUIC.swift
adam-fowler: Can we remove this

PR #267, opened 2026-07-20
Default to `Logger.current` in public API

fpseverino:
- Use by default the task-local logger in `MQTTConnection.withConnection` and `MQTTSession.init`
- Add a custom scoping trait to tests to automatically set our custom logger in all tests
- Update documentation

### DIFF
--- a/Tests/MQTTNIOTests/Utils/Logger+withLogLevel.swift
+++ b/Tests/MQTTNIOTests/Utils/Logger+withLogLevel.swift
@@ -18,6 +18,8 @@ extension Logger {
 }
 
 struct DefaultLoggerTrait: TestTrait, SuiteTrait, TestScoping {
+    let logLevel: Logger.Level
+
     var isRecursive: Bool { true }
 
     func provideScope(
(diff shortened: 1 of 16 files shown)


Tests/MQTTNIOTests/Utils/Logger+withLogLevel.swift
```diff
@@ -15,3 +16,24 @@ extension Logger {
         return logger
     }
 }
+
+struct DefaultLoggerTrait: TestTrait, SuiteTrait, TestScoping {
```
adam-fowler, line 20: Perhaps the default logger should provide a way to set the logger level

adam-fowler (commented):
One minor thing, otherwise looks good

Tests/MQTTNIOTests/Utils/Logger+withLogLevel.swift
fpseverino: Yeah that would be nice, I just added it

adam-fowler (approved):

PR #270, opened 2026-08-04
Prevent MQTT task timeout double completion

jollyjinx:
Fixes a race where an MQTT request timeout could complete its promise without removing the request from the channel handler's state machine.

If the connection subsequently closed or failed, the retained task was failed a second time. For async callers this could trigger:

> SWIFT TASK CONTINUATION MISUSE: sendPacket(_:checkInbound:) tried to resume its continuation more than once

## Changes

- Route request timeouts through `MQTTChannelHandler.cancel`.
- Remove the timed-out task from the state machine before failing its promise.
- Use `MQTTTask.fail` for cancellation so its scheduled timeout is cancelled consistently.
- Add a regression test covering a QoS 1 publish that times out immediately before the connection closes.

This ensures that timeout, cancellation, connection failure, and response handling all compete through the same task-ownership path, allowing only one of them to complete the promise.

## Testing

- Added `publishTimeoutThenConnectionClose`.
- Confirmed that the new test reproduces the continuation double-resume crash before the fix.
- Confirmed that the test passes after the fix.

### DIFF
--- a/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
+++ b/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler.swift
@@ -169,7 +169,7 @@ final class MQTTChannelHandler: ChannelDuplexHandler {
         self.eventLoop.assertInEventLoop()
         switch self.stateMachine.cancel(requestID: requestID) {
         case .failTask(let cancelledTask):
-            cancelledTask.promise.fail(error)
+            cancelledTask.fail(error)
         case .doNothing:
             break
         }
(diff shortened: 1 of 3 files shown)


adam-fowler:
Thanks for this and diagnosing the error. While your solution does work, I think there should probably be a structural change where we deal with timeout at the state machine level. We were also thinking of moving to using a `NIOScheduledCallbackHandler` instead of `scheduleTask(_:)`.

jollyjinx:
sounds like it will be the right solution. Looking forward to test it with my mqttclients.

adam-fowler:
I think @fpseverino is going to supply a solution

adam-fowler:
Closed in favour of #271

PR #271, opened 2026-08-07
Switch timeout from `scheduleTask` to `NIOScheduledCallbackHandler`

fpseverino:
- Switches the implementation of the `MQTTTask` timeout from `scheduleTask` to `NIOScheduledCallbackHandler`
- Fixes the bug reported in #270 by @jollyjinx

### DIFF
--- a/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
+++ b/Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
@@ -248,7 +248,7 @@ extension MQTTChannelHandler {
 
         @usableFromInline
         enum HitDeadlineAction {
-            case cancelTask(requestID: Int)
+            case cancelTasks(requestIDs: [Int])
             case reschedule(NIODeadline)
             case clearCallback
         }
@@ -259,9 +259,12 @@ extension MQTTChannelHandler {
             case .uninitialized:
                 preconditionFailure("Cannot cancel when uninitialized")
             case .initialized(let state):
-                if let timedOutTask = state.tasks.tasks.lazy.min(by: { $0.deadline < $1.deadline }), timedOutTask.deadline <= now {
+                let timedOutRequestIDs = state.tasks.tasks.lazy
+                    .filter { $0.deadline <= now }
+                    .map { $0.requestID }
+                if !timedOutRequestIDs.isEmpty {
                     self = .initialized(state)
-                    return .cancelTask(requestID: timedOutTask.requestID)
+                    return .cancelTasks(requestIDs: .init(timedOutRequestIDs))
                 }
                 if let earliestDeadline = state.tasks.tasks.lazy.map({ $0.deadline }).min() {
                     self = .initialized(state)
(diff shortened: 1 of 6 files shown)


Sources/MQTTNIO/ChannelHandlers/MQTTTask.swift
adam-fowler: Does this still need to be a class?

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
```diff
@@ -72,7 +74,17 @@ extension MQTTChannelHandler {
                         // should this task respond to inbound packet
                         if try task.checkInbound(packet) {
                             self.remove(at: index)
-                            return .succeedTask(task)
+                            let deadlineCallback: DeadlineCallbackAction =
+                                if self.tasks.isEmpty {
+                                    .cancel
+                                } else {
+                                    if let earliestDeadline = self.tasks.map({ $0.deadline }).min() {
```
adam-fowler:
```suggestion
                                    if let earliestDeadline = self.tasks.lazy.map({ $0.deadline }).min() {
```

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
```diff
@@ -227,6 +246,35 @@ extension MQTTChannelHandler {
             }
         }
 
+        @usableFromInline
+        enum HitDeadlineAction {
+            case cancelTask(requestID: Int)
```
adam-fowler: Is there a chance we might have to cancel multiple tasks?

adam-fowler (commented):
Looking good. I've added a few comments

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
fpseverino: I've thought a bit about this: that two request are made on the exact same instant is unlikely, but the NIO scheduled callback could be imprecise and be called after more than one request timed-out, but honestly I don't know. What are your thoughts on this?

adam-fowler: I think we're better being cautious. So I'd allow for cancelling multiple tasks.

adam-fowler (changes requested):
Forgot to request one other thing. Can we add some state machine tests as well please

fpseverino:
> Forgot to request one other thing. Can we add some state machine tests as well please

Sure, I'll start working on it and open another PR 👍🏻

Sources/MQTTNIO/ChannelHandlers/MQTTChannelHandler+stateMachine.swift
fpseverino: I implemented it in [my last commit](https://github.com/swift-server-community/mqtt-nio/pull/271/commits/0b5e1cacb0d13d84d8bfa2c443a4a16ecefb4204), WDYT?
The test I added in the commit was also successful before, but I saw via some debug `print`s that before the callback was called twice, and now only once.

adam-fowler (approved):
Ok this looks good.
Are you planning to do the state machine tests in a separate PR?

fpseverino:
> Are you planning to do the state machine tests in a separate PR?

Yes I am, I hope to do it soon

adam-fowler:
ok I'll merge

PR #273, opened 2026-08-14
Tracing: propagate trace id via MQTTPublishInfo user properties

### LINKED ISSUE #214: v3: Swift Distributed Tracing
Provide built-in support for distributed tracing via [Swift Distributed Tracing](https://github.com/apple/swift-distributed-tracing).

adam-fowler:
Add support for tracing. 
Publish spans are created automatically as long as you have set a tracer.
The trace context is then propagated via MQTTPublishInfo user properties
Can't easily create spans for subscriptions automatically, so added a function `MQTTConnection.withMessageSpan()` to extract trace context from MQTTPublishInfo

So subscriptions would work like this
```swift
for try await message in subscription {
    try await connection.withMessageSpan(message) { span in
        doStuff()
    }
}
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,7 +29,12 @@ let package = Package(
     products: [
         .library(name: "MQTTNIO", targets: ["MQTTNIO"])
     ],
+    traits: [
+        .trait(name: "DistributedTracingSupport"),
+        .default(enabledTraits: ["DistributedTracingSupport"]),
+    ],
     dependencies: [
+        .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.4.0"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.14.0"),
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.100.0"),
         .package(url: "https://github.com/apple/swift-nio-ssl.git", from: "2.36.0"),
@@ -43,6 +48,7 @@ let package = Package(
             name: "MQTTNIO",
             dependencies: [
                 .product(name: "Logging", package: "swift-log"),
+                .product(name: "Tracing", package: "swift-distributed-tracing", condition: .when(traits: ["DistributedTracingSupport"])),
                 .product(name: "NIO", package: "swift-nio"),
                 .product(name: "NIOHTTP1", package: "swift-nio"),
                 .product(name: "NIOWebSocket", package: "swift-nio"),
@@ -66,6 +72,7 @@ let package = Package(
             dependencies: [
                 .target(name: "MQTTNIO"),
                 .product(name: "InMemoryLogging", package: "swift-log"),
+                .product(name: "InMemoryTracing", package: "swift-distributed-tracing", condition: .when(traits: ["DistributedTracingSupport"])),
                 .product(name: "NIOFoundationCompat", package: "swift-nio"),
             ],
             swiftSettings: defaultSwiftSettings
--- a/Package.swift
+++ b/Package.swift
@@ -30,8 +30,8 @@ let package = Package(
         .library(name: "MQTTNIO", targets: ["MQTTNIO"])
     ],
     traits: [
-        .trait(name: "DistributedTracingSupport"),
-        .default(enabledTraits: ["DistributedTracingSupport"]),
+        .trait(name: "DistributedTracing"),
+        .default(enabledTraits: ["DistributedTracing"]),
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.4.0"),
@@ -48,7 +48,7 @@ let package = Package(
             name: "MQTTNIO",
             dependencies: [
                 .product(name: "Logging", package: "swift-log"),
-                .product(name: "Tracing", package: "swift-distributed-tracing", condition: .when(traits: ["DistributedTracingSupport"])),
+                .product(name: "Tracing", package: "swift-distributed-tracing", condition: .when(traits: ["DistributedTracing"])),
                 .product(name: "NIO", package: "swift-nio"),
                 .product(name: "NIOHTTP1", package: "swift-nio"),
                 .product(name: "NIOWebSocket", package: "swift-nio"),
@@ -72,7 +72,7 @@ let package = Package(
             dependencies: [
                 .target(name: "MQTTNIO"),
                 .product(name: "InMemoryLogging", package: "swift-log"),
-                .product(name: "InMemoryTracing", package: "swift-distributed-tracing", condition: .when(traits: ["DistributedTracingSupport"])),
+                .product(name: "InMemoryTracing", package: "swift-distributed-tracing", condition: .when(traits: ["DistributedTracing"])),
                 .product(name: "NIOFoundationCompat", package: "swift-nio"),
             ],
             swiftSettings: defaultSwiftSettings
--- a/Sources/MQTTNIO/Connection/MQTTConnection.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnection.swift
@@ -226,7 +226,7 @@ public final actor MQTTConnection: Sendable {
         #if DistributedTracingSupport
         var info = MQTTPublishInfo(qos: qos, retain: retain, dup: false, topicName: topicName, payload: payload, properties: .init())
         if self.configuration.version == .v5_0, let tracer = self.tracer {
-            try await tracer.withSpan("PUBLISH", ofKind: .producer) { span in
+            try await tracer.withSpan("PUBLISH \(topicName)", ofKind: .producer) { span in
                 if !(span is NoOpTracer.Span) {
                     tracer.inject(span.context, into: &info, using: self.configuration.tracing.contextPropagator.injector)
                     span.updateAttributes { attributes in
--- a/Sources/MQTTNIO/Connection/MQTTConnection.swift
+++ b/Sources/MQTTNIO/Connection/MQTTConnection.swift
@@ -230,7 +230,7 @@ public final actor MQTTConnection: Sendable {
             defer { span?.end() }
 
             if let span, !(span is NoOpTracer.Span) {
-                InstrumentationSystem.instrument.inject(span.context, into: &info, using: self.configuration.tracing.contextPropagator.injector)
+                self.tracer?.inject(span.context, into: &info, using: self.configuration.tracing.contextPropagator.injector)
                 span.updateAttributes { attributes in
                     self.applyCommonPublishAttributes(to: &attributes)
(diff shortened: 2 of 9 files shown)


Package.swift
```diff
@@ -29,7 +29,12 @@ let package = Package(
     products: [
         .library(name: "MQTTNIO", targets: ["MQTTNIO"])
     ],
+    traits: [
+        .trait(name: "DistributedTracingSupport"),
```
fpseverino: I think that nowadays the best-practice is to drop the "Support" suffix, following the style of Swift Configuration
```suggestion
        .trait(name: "DistributedTracing"),
```

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -256,6 +275,71 @@ public final actor MQTTConnection: Sendable {
...
+            configuration.tracing.attributeNames.messagingOperationName: "publish",
+        ]
+        if let remoteAddress = channel.remoteAddress {
+            commonAttributes[configuration.tracing.attributeNames.networkPeerAddress] = remoteAddress.ipAddress
+            commonAttributes[configuration.tracing.attributeNames.networkPeerPort] = remoteAddress.port
+        }
+        switch address?.value {
+        case let .hostname(host, port):
+            commonAttributes[configuration.tracing.attributeNames.serverAddress] = host
+            commonAttributes[configuration.tracing.attributeNames.serverPort] = (port == 6379 ? nil : port)
```
```suggestion
            commonAttributes[configuration.tracing.attributeNames.serverPort] = (port == 1883 ? nil : port)
```

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -256,6 +275,71 @@ public final actor MQTTConnection: Sendable {
...
+            configuration.tracing.attributeNames.messagingOperationName: "subscribe",
+        ]
+        if let remoteAddress = channel.remoteAddress {
+            commonAttributes[configuration.tracing.attributeNames.networkPeerAddress] = remoteAddress.ipAddress
+            commonAttributes[configuration.tracing.attributeNames.networkPeerPort] = remoteAddress.port
+        }
+        switch address?.value {
+        case let .hostname(host, port):
+            commonAttributes[configuration.tracing.attributeNames.serverAddress] = host
+            commonAttributes[configuration.tracing.attributeNames.serverPort] = (port == 6379 ? nil : port)
```
```suggestion
            commonAttributes[configuration.tracing.attributeNames.serverPort] = (port == 1883 ? nil : port)
```

Sources/MQTTNIO/Connection/MQTTConnection.swift
```diff
@@ -256,6 +275,71 @@ public final actor MQTTConnection: Sendable {
...
+        return commonAttributes
+    }
+
+    @usableFromInline
+    nonisolated func applyCommonSubscribeAttributes(to attributes: inout SpanAttributes) {
+        attributes.merge(self.commonSubscribeSpanAttributes)
+    }
+
+    #endif
+
```
```suggestion
    @usableFromInline
    nonisolated func applyCommonSubscribeAttributes(to attributes: inout SpanAttributes) {
        attributes.merge(self.commonSubscribeSpanAttributes)
    }
    #endif

```

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
```diff
@@ -255,6 +259,12 @@ public struct MQTTConnectionConfiguration: Sendable {
     /// Transport to use for the connection and its associated configuration.
     public var transport: Transport
 
+    #if DistributedTracingSupport
+    /// The distributed tracing configuration to use for this connection.
+    /// Defaults to using the globally bootstrapped tracer with OpenTelemetry semantic conventions.
```
fpseverino, line 264: Are there OpenTelemetry semantic conventions for MQTT?

Perhaps we should mention they are the conventions from HiveMQ/EMQX

Sources/MQTTNIO/Tracing.swift
```diff
@@ -0,0 +1,115 @@
...
+/// A configuration object that defines distributed tracing behavior of a MQTT client.
+public struct MQTTTracingConfiguration: Sendable {
+    /// The tracer to use, or `nil` to disable tracing.
+    /// Defaults to the globally bootstrapped tracer.
+    public var tracer: (any Tracer)? = InstrumentationSystem.tracer
+
+    /// Tracing context propagator
+    public var contextPropagator: any MQTTContextPropagator = .userProperties
+
+    /// The attribute names used in spans created by Valkey. Defaults to OpenTelemetry semantics.
```
fpseverino: Look out for "Valkey", which appears a few times in this file.
Also, the same question as before regarding OpenTelemetry semantics.

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
fpseverino, line 1: Shall we add to the `ConfigReader` initializer DocC comment that it is impossible to get the tracing configuration with Swift Configuration, and that the user should set it manually?

fpseverino (changes requested):
This is great, we should also add a CI job that runs the tests with all default traits disabled, something like [this](https://github.com/fpseverino/stomp-nio/blob/cf0f599a655a07c68c938296c5e2392187123ce4/.github/workflows/ci.yml#L113-L141).

Moreover, I noticed the CI fail on Linux with Swift 6.3 before, is it perhaps related to [this](https://github.com/swift-server-community/mqtt-nio/issues/274)?

Sources/MQTTNIO/Connection/MQTTConnection.swift
adam-fowler: what's the difference here?

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
adam-fowler: https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
adam-fowler: At this point I think we should leave the ConfigReader initialiser

Sources/MQTTNIO/Connection/MQTTConnection.swift
fpseverino: It's really just a nit, there's an empty line between the curly brace at the end of the function and the `#endif`, but there isn't one under `#if DistributedTracing`, so my suggestion is to remove it at the end to keep it consistent

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
fpseverino: Ohhh I see 👍🏻

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
fpseverino: What do you mean?

adam-fowler: I meant not to bother at the moment. I don't think the trace names need to be configurable via Swift-configuration.

adam-fowler:
@fpseverino  I added an option to link subscribe (process) spans to publish (creation) spans via a span link. You can change this to make the process span a child of the publish span. See https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#message-creation-context-as-parent-of-process-span for more details on linking creation and process spans. You can use config option `createChildConsumerSpans` to control which method is chosen (It defaults to using a span link as per OTel docs).

This option is only ever used in the `withMessageSpan` function. What are your thoughts on making it a parameter of the function instead?

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
```diff
@@ -255,6 +259,12 @@ public struct MQTTConnectionConfiguration: Sendable {
     /// Transport to use for the connection and its associated configuration.
     public var transport: Transport
 
+    #if DistributedTracing
+    /// The distributed tracing configuration to use for this connection.
+    /// Defaults to using the globally bootstrapped tracer with OpenTelemetry semantic conventions.
+    public var tracing: MQTTTracingConfiguration = .init()
```
fpseverino, line 265: Shall we also add another initializer that has a `tracing` parameter?

Tests/MQTTNIOTests/MQTTConnectionTests.swift
```diff
@@ -6,11 +6,13 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+import InMemoryTracing
 import Logging
 import NIOCore
 import NIOEmbedded
 import NIOHTTP1
 import Testing
+import Tracing
```
fpseverino: These imports should be guarded by `#if DistributedTracing`. A CI check that looks something like [this](https://github.com/fpseverino/stomp-nio/blob/c69d3d5c9d593d3a8b366252200a9bbb0c25a5b2/.github/workflows/ci.yml#L113-L141) could help us spot these things

Sources/MQTTNIO/Tracing.swift
```diff
@@ -0,0 +1,152 @@
...
+    public var tracer: (any Tracer)?
+
+    /// Tracing context propagator
+    public var contextPropagator: any MQTTContextPropagator
+
+    /// Controls how publish and subscribe spans are linked. By default the subscribe
+    /// span is linked to the publish context and is a child of current span.
+    /// If this is set to true the subscribe span is set to be a child of the publish
+    /// context and a link to the current span is stored.
+    public var createChildConsumerSpans: Bool
```
fpseverino: > What are your thoughts on making it a parameter of the function instead?

Yes absolutely, that way the user could customize the behavior for every subscription, instead of the same for the whole connection, right?

fpseverino (approved):
I'm not an expert, but this looks really good! We really need to update the DocC catalog though, perhaps with an article on the matter

Tests/MQTTNIOTests/MQTTConnectionTests.swift
adam-fowler: Added CI for this, needed to add --build-tests to ensure we catch this in tests

Sources/MQTTNIO/Connection/MQTTConnectionConfiguration.swift
adam-fowler: No I'm happy to leave as

adam-fowler:
> I'm not an expert, but this looks really good! We really need to update the DocC catalog though, perhaps with an article on the matter

I'll do a tracing article in a separate PR

